Issue #8, opened 2018-03-05
Use convolutional filters of different sizes


ericagreene:
Apparently it is common practice to use convolutional filters of different sizes, but the CNN implementation in this repo does not do this. So we should add it.

PR #9, opened 2018-03-07
Add filters of varying sizes and dropout to CNN

ericagreene:
This change is a rewrite of the CNN model which adds
* filters of multiple sizes
* dropout during training

 I had previously adapted the implementation from one of the examples in the TensorFlow Github repo ([here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/learn/text_classification_cnn.py)), but JJ pointed out that it only used one filter size. I also found the `tf.layers` API a little confusing. 

JJ pointed me to [this](http://www.wildml.com/2015/12/implementing-a-cnn-for-text-classification-in-tensorflow/) Wild ML CNN tutorial that uses filters of multiple sizes and adds dropout during training. It also uses lower level TensorFlow API's and is a bit easier to read. 

I ran both models on the Kaggle data to predict toxicity and the AUC went up from .76 to .80. Without any tuning, just run for the same number of time steps.

The update to `run.sh` makes it possible to run multiple models remotely at the same time. Previously we were saving all the checkpoints to the same Cloud Storage directory. 

This closes https://github.com/conversationai/conversationai-models/issues/8

### DIFF
--- a/kaggle-classification/trainer/model.py
+++ b/kaggle-classification/trainer/model.py
@@ -42,15 +42,13 @@ DATA_SEED = 48173 # Random seed used for splitting the data into train/test
 MAX_LABEL = 2
 MAX_DOCUMENT_LENGTH = 500 # Max length of each comment in words
 
+# CNN parameters
 CNNParams = namedtuple(
-  'CNNParams', ['WINDOW_SIZE', 'EMBEDDING_SIZE','POOLING_WINDOW', 'POOLING_STRIDE',
-                'N_FILTERS', 'FILTER_SHAPE1', 'FILTER_SHAPE2'])
-cnn_values = {'WINDOW_SIZE':20, 'EMBEDDING_SIZE':20, 'POOLING_WINDOW':4,
-              'POOLING_STRIDE':2, 'N_FILTERS':10}
-cnn_values['FILTER_SHAPE1'] = [cnn_values['WINDOW_SIZE'], cnn_values['EMBEDDING_SIZE']]
-cnn_values['FILTER_SHAPE2'] = [cnn_values['WINDOW_SIZE'], cnn_values['N_FILTERS']]
-CNN_PARAMS = CNNParams(**cnn_values)
+  'CNNParams',['EMBEDDING_SIZE','N_FILTERS', 'FILTER_SIZES', 'DROPOUT_KEEP_PROB'])
+CNN_PARAMS = CNNParams(
+  EMBEDDING_SIZE=20, N_FILTERS=10, FILTER_SIZES=[2,3,4,5],  DROPOUT_KEEP_PROB=.75)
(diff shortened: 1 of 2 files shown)


kaggle-classification/trainer/model.py
```diff
@@ -42,15 +42,13 @@
...
-  'CNNParams', ['WINDOW_SIZE', 'EMBEDDING_SIZE','POOLING_WINDOW', 'POOLING_STRIDE',
-                'N_FILTERS', 'FILTER_SHAPE1', 'FILTER_SHAPE2'])
-cnn_values = {'WINDOW_SIZE':20, 'EMBEDDING_SIZE':20, 'POOLING_WINDOW':4,
-              'POOLING_STRIDE':2, 'N_FILTERS':10}
-cnn_values['FILTER_SHAPE1'] = [cnn_values['WINDOW_SIZE'], cnn_values['EMBEDDING_SIZE']]
-cnn_values['FILTER_SHAPE2'] = [cnn_values['WINDOW_SIZE'], cnn_values['N_FILTERS']]
-CNN_PARAMS = CNNParams(**cnn_values)
+  'CNNParams',['EMBEDDING_SIZE','N_FILTERS', 'FILTER_SIZES', 'DROPOUT_KEEP_PROB'])
+CNN_PARAMS = CNNParams(
+  EMBEDDING_SIZE=20, N_FILTERS=10, FILTER_SIZES=[2,3,4,5],  DROPOUT_KEEP_PROB=.75)
```
nthain: This seems like quite a small embedding size. As you continue to experiment with parameters, I would try adjusting the embedding size up to 50-100 which is more standard.

kaggle-classification/trainer/model.py
```diff
@@ -144,56 +142,68 @@ def estimator_spec_for_softmax_classification(logits, labels, mode):
...
+    h_drop = tf.nn.dropout(h_pool_flat, 1.0)
+
+    if mode == tf.estimator.ModeKeys.TRAIN:
+      h_drop = tf.nn.dropout(h_pool_flat, dropout_keep_prob)
+
+  # Add a fully connected layer to do prediction
+  with tf.name_scope("output"):
+    W = tf.Variable(tf.truncated_normal([num_filters_total, MAX_LABEL], stddev=0.1), name="W")
+    b = tf.Variable(tf.constant(0.1, shape=[MAX_LABEL]), name="b")
+    scores = tf.nn.xw_plus_b(h_drop, W, b, name="scores")
```
nthain: This works. You can do it in fewer lines of code with tf.layers.dense.

ericagreene: Ok, good to know!

ericagreene:
Thanks for reviewing @nthain!

PR #11, opened 2018-03-09
Initial draft of attention codelab

nthain:
There is still a bunch I'd like to add or improve, but this should function as a standalone first draft of the codelab.


attention-codelab/process_figshare.py
```diff
@@ -0,0 +1,44 @@
+from __future__ import absolute_import
```
iislucas: Always add apache 2 license info to top. 

Add top level docstring/comment also to explain what this code does.

nthain: Done. Please check if the license information is in the format you expected.

iislucas:
It's not easy to comment on locations in the ipynb file, so you get them in this comment :) 

@ Load & Explore Data:
- Given we have a python script already to load the data, why not have it also do the downloading? Avoid extra steps. 
- Why does the rev_id contain a decimal point? Better to treat it as a string I think... ?

@ Hyperparameters:
 - Add description for the hyperparams

@ Step 0: Text Preprocessing: 
 - Add link to ToensorFlow tokenizers in the intro about other tokenizers.
 - Add link to VocabularyProcessor http://tflearn.org/data_utils/#vocabulary-processor 
 - The comment "# Train the vocab_processor from the training set" leaves the reader a little unsure what is happening, the details are then given below. Only in the section after does it become apparent why this is there, and what it might mean. I suggest instead of leading with code, lead with the description, and then show the code. Same for embed definition. You show simple code snippet, but reader currently lacks context. Leading/adding a comment like: "Using these embeddings, we define a  simple helper function to lookup the word and compute its embedding. You should also say what happens when a word is not in the dictionary/embedding table. 

Generally, I think it's good to link to terms, e.g. gated recurrent units.

Nit: wrap lines to make it more readable. e.g. in [9]. 

@ Encode : 
 - I think the diagram and the following code are inconsistent: the diagram shows a single output from the last cell, but in the code it looks like it's giving an array output? Maybe the figure is in the wrong place and should be moved down to where you talk about it? 
 - The diagram is for one-directional attention: but we don't say what we do to deal with two directions. Write a bit more about that. Is the idea that each direction has its own attention that gets applied to it's part of the vector? This seems like a key architectural decision. 

@ Predict: 
 - link to "estimator framework". 
 - Clarify auc is ROC-AUC (and link?)
 - I think relationship between alphas and logits should be specified. Not clear to me what's happening here. I think we can (and should) make this understandable even if you don't understand estimator framework, but explaining what the parameters are and how they relate. 

General things: 
 * To get more detailed feedback and language suggestions, maybe copy/paste notebook into a google doc? 
 * Use python 3 type hints to make code more readable.

nthain:
Thanks for the review Lucas! I've just completed all of the changes. Your suggestions are copied below and my responses are inline. PTAL.

@ Load & Explore Data:
* Given we have a python script already to load the data, why not have it also do the downloading? Avoid extra steps. - DONE
* Why does the rev_id contain a decimal point? Better to treat it as a string I think... ? - DONE

@ Hyperparameters:
* Add description for the hyperparams - DONE

@ Step 0: Text Preprocessing:

* Add link to ToensorFlow tokenizers in the intro about other tokenizers. - DONE
* Add link to VocabularyProcessor http://tflearn.org/data_utils/#vocabulary-processor - DONE
* The comment "# Train the vocab_processor from the training set" leaves the reader a little unsure what is happening, the details are then given below. Only in the section after does it become apparent why this is there, and what it might mean. I suggest instead of leading with code, lead with the description, and then show the code. Same for embed definition. You show simple code snippet, but reader currently lacks context. Leading/adding a comment like: "Using these embeddings, we define a simple helper function to lookup the word and compute its embedding. You should also say what happens when a word is not in the dictionary/embedding table. - DONE

* Generally, I think it's good to link to terms, e.g. gated recurrent units. - DONE

* Nit: wrap lines to make it more readable. e.g. in [9]. - DONE

@ Encode :

* I think the diagram and the following code are inconsistent: the diagram shows a single output from the last cell, but in the code it looks like it's giving an array output? Maybe the figure is in the wrong place and should be moved down to where you talk about it?
 - Added a second diagram to clarify
* The diagram is for one-directional attention: but we don't say what we do to deal with two directions. Write a bit more about that. Is the idea that each direction has its own attention that gets applied to it's part of the vector? This seems like a key architectural decision.
 - Added another diagram to clarify

@ Predict:

* link to "estimator framework" - DONE
* Clarify auc is ROC-AUC (and link?) - DONE
* I think relationship between alphas and logits should be specified. Not clear to me what's happening here. I think we can (and should) make this understandable even if you don't understand estimator framework, but explaining what the parameters are and how they relate. - I think this is explained in the final paragraph of Step 4. Let me know if it's still unclear.

General things:

* To get more detailed feedback and language suggestions, maybe copy/paste notebook into a google doc? - DONE. You can leave suggestions at the doc here: https://docs.google.com/document/d/1qL7Ld1_ijwF_lvyI1T63HhEShlr8-v-XJnsEK540rXQ/edit?usp=sharing

* Use python 3 type hints to make code more readable. - I decided not to do this because I'd like the notebook to be accessible to those that don't use type hints.

iislucas:
Thanks Nithum! I'll take another look shortly. 

w.r.t. type hints: others don't have to use the type-hints you provide. They are still useful to readers of the code and don't affect runtime or compatibility.

attention-codelab/process_figshare.py
sorensenjs: I think you just need to swap the license and the comment.

attention-codelab/process_figshare.py
```diff
@@ -0,0 +1,107 @@
...
+        if not os.path.isfile(os.path.join(download_data_dir, file)):
+            already_exist = False
+    
+    if already_exist:
+        print('Figshare data already exists.')
+        return
+    
+    print('Downloading toxicity_annotations.tsv...', end = '')
+    urlretrieve('https://ndownloader.figshare.com/files/7394539', 
+                os.path.join(download_data_dir, 'toxicity_annotations.tsv'))
```
sorensenjs: why don't you do this in the loop above
if not os.path.isfile(...):
  urlretreive(..., os.path.join(file))
also pull ndownload.figshare.com prefix to a constant string somewhere.

nthain: Good suggestion! Done.

attention-codelab/process_figshare.py
```diff
@@ -0,0 +1,107 @@
...
+                                              sep = '\t', 
+                                              dtypes = {'rev_id': 'str'})
+    toxicity_annotations = pd.read_csv(os.path.join(input_data_dir, 'toxicity_annotations.tsv'), 
+                                       sep = '\t',
+                                       dtypes = {'rev_id': 'str'})
+
+    annotations_gped = toxicity_annotations.groupby('rev_id', as_index=False).agg({'toxicity': 'mean'})
+    all_data = pd.merge(annotations_gped, toxicity_annotated_comments, on = 'rev_id')
+
+    all_data['comment'] = all_data['comment'].apply(lambda x: x.replace("NEWLINE_TOKEN", " "))
```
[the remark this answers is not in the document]
nthain: Ah, that's much better. Done!

PR #14, opened 2018-03-12
Convert contrib to tflearn VocabularyProcessor

sorensenjs:
tf.contrib is deprecated. switch to tflearn


sorensenjs:
Updated and tested that ran both locally and remotely.

PR #17, opened 2018-03-13
support python 3, readme tweaks, and run_local fixes

iislucas:
* Moved local data to a `local_data` directory. 
* Added command line instructions for downloading local data. 
* Made `run_local.sh` use data in `local_data` directory (so it's really totally local)
* Fixed loading code to work in python3 (utf-8 fixes)
* Adds more detailed instructions, and recommends using virtualenv
* Sort the `requirements.txt` because sorting is cool.

### DIFF
--- a/kaggle-classification/.gitignore
+++ b/kaggle-classification/.gitignore
@@ -1,6 +1,3 @@
-# Don't version control data
-*.csv
-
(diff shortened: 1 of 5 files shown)


ericagreene:
Looks good, thanks for the updates!

PR #18, opened 2018-03-13
Introduce Keras model

jjtan:


kaggle-classification/keras-trainer/model.py
```diff
@@ -0,0 +1,184 @@
...
+  def _load_model(self):
+    try:
+      self.model = load_model(self.model_path)
+      print('Model loaded from: {}'.format(self.model_path))
+    except IOError:
+      print('Could not load model at: {}'.format(self.model_path))
+
+  def _setup_tokenizer(self):
+    words = []
+    with open(self.embeddings_path) as f:
```
nthain: Is there a way to avoid reading this file twice?

kaggle-classification/keras-trainer/model.py
```diff
@@ -0,0 +1,184 @@
...
+        self.hparams.embedding_dim,
+        weights=[self.embedding_matrix],
+        input_length=self.hparams.max_sequence_length,
+        trainable=False)(I)
+    X5 = Conv1D(128, 5, activation='relu', padding='same')(E)
+    X5 = MaxPooling1D(250, padding='same')(X5)
+    X4 = Conv1D(128, 4, activation='relu', padding='same')(E)
+    X4 = MaxPooling1D(250, padding='same')(X4)
+    X3 = Conv1D(128, 3, activation='relu', padding='same')(E)
+    X3 = MaxPooling1D(250, padding='same')(X3)
```
nthain: In the future, might want to use GlobalMaxPooling1D for the final layer.

kaggle-classification/keras-trainer/model.py
jjtan: Let me defer. I don't think it's simple because the tokenizer needs to be fully setup because we created the embedding matrix.

PR #21, opened 2018-03-13
Ngrams1

sorensenjs:
Creates character trigrams instead of a vocabulary.


sorensenjs:
Creates a "character" model by iterating over trigrams of characters instead of words.

Issue #10, opened 2018-03-09
Configure ml-engine to use GPUs


ericagreene:
This should be a small code change, but involves reading the documentation on how to set this up in GCP.  https://cloud.google.com/ml-engine/docs/using-gpus

It would also be great to add a flag to the python script to switch between GPU's and CPU's

PR #23, opened 2018-03-13
Use GPUs by cloud jobs

iislucas:
Also updates cloud job names to include username. Adds a few comments we'll eventually want to remove, but are helpful for now while we all explore this.


ericagreene:
Closes https://github.com/conversationai/conversationai-models/issues/10

PR #28, opened 2018-03-13
Changing default hyperparameters and number of run steps

nthain:
The new config file kicks off 100 trials of 5000 steps each (although early stopping is allowed).


kaggle-classification/hparam_config.yaml
```diff
@@ -22,10 +23,15 @@ trainingInput:
...
         minValue: 0.5
         maxValue: 1
-        scaleType: UNIT_LINEAR_SCALE
\ No newline at end of file
+        scaleType: UNIT_LINEAR_SCALE
+      - parameterName: learning_rate
+        type: DOUBLE
+        minValue: 0.000001
+        maxValue: 0.1
+        scaleType: UNIT_LOG_SCALE
```
ericagreene: Can you add a newline to the end of the file?

nthain: Added.

nthain:
Thanks Erica, PTAL (I didn't get approval last time so didn't want to submit yet).

ericagreene:
We talked about this offline, but it would be great if the default behavior for training a model on ml-engine remotely will just train one model, not do parameter tuning. I think that should be a small change, but it's pretty important if we don't want to churn through 2 hours of GPU resources everytime we kick off a job.

ericagreene:
We talked offline and decided: 
* add a default config file for just GPUs
* add a flag to the shell run script which can take an optional non-default config with hyper-parameter tuning (or other things)

nthain:
Done. You can now use the -c flag to specify a config file. To train with hyperparameters, just use:
./bin/run -c hparam_config.yaml

This is included in the comments of the run file and also can be found by using the '-h' flag for help.

nthain:
PTAL

ericagreene:
Approved! 

Do you mind rebasing and adding newlines to the end of the files? Then feel free to merge.

PR #31, opened 2018-03-14
Kaggle keras init

jjtan:


kaggle-classification/keras-trainer/model.py
```diff
@@ -70,7 +77,7 @@ def train(self, train):
 
     callbacks = [
         ModelCheckpoint(
-            self.model_path, save_best_only=True, verbose=True),
+            TEMPORARY_MODEL_PATH, save_best_only=True, verbose=True),
```
nthain: Please clean up these files once you're done with them to avoid having to manually delete them when running locally.

kaggle-classification/run_keras.sh
```diff
@@ -0,0 +1,36 @@
...
+#
+# Setup Steps:
+# 1. Install the gcloud SDK
+# 2. Authenticate with the GCP project you want to use, `gcloud config set project [my-project]`
+# 3. Put the train and test data in Cloud Storage, `gsutil cp [DATA_FILE] gs://[BUCKET_NAME]/`
+#
+
+# Edit these!
+BUCKET_NAME=kaggle-model-experiments
+JOB_NAME=test_kaggle_training
```
nthain: Replace test with ${USER}

PR #35, opened 2018-03-15
Initial attempt at attention and adding model type as a hyperparameter.

nthain:


kaggle-classification/keras_trainer/model.py
```diff
@@ -74,7 +84,7 @@ def train(self, train):
             monitor='val_loss', mode='auto')
     ]
 
-    model.fit(
+    self.model.fit(
```
jjtan: should this be model.fit?

PR #36, opened 2018-03-15
Convolutions as lists, deepening attention, and adding Tensorboard

nthain:


ericagreene:
Very cool! Can you add a description to the PR and include the command for how to run this?

nthain:
Thanks for the feedback Erica! I'll actually add the command for running tensorboard in a separate PR (as once I got started, I ended up doing some bigger refactoring).

Issue #42, opened 2018-03-26
Support Hyperparameter Tuning in Keras


nthain:
Implement hyperparameter tuning in Keras

ericagreene:
@jjtan This is done, right? Want to be the one that closes it? :)

PR #47, opened 2018-03-30
Kaggle keras init

jjtan:
Hyper param tuning in ml engine for keras

Should be merged after #46 
Issue #42


kaggle-classification/hparam_config.yaml
```diff
@@ -1,5 +1,6 @@
 trainingInput:
   ## BASIC_GPU uses single NVIDIA Tesla K80 GPU.
+  pythonVersion: '3.5'
```
ericagreene: Is this python version required for hyperparameter tuning?

kaggle-classification/keras_trainer/model.py
```diff
@@ -42,7 +43,7 @@
     learning_rate=0.00005,
     dropout_rate=0.5,
     batch_size=128,
-    epochs=1,
+    epochs=10,
```
ericagreene: Is there a reason this changed back to 10?

kaggle-classification/keras_hparam_config.yaml
```diff
@@ -9,3 +9,28 @@ trainingInput:
   # parameterServerType: large_model
   # workerCount: 9
   # parameterServerCount: 3
+  hyperparameters:
```
ericagreene: Can you add a comment indicating roughly how many versions of the model will be run with this config? It'd be nice to know if I'm setting off 10 jobs or 100 jobs or 1000 jobs.

kaggle-classification/keras_trainer/model.py
```diff
@@ -98,7 +99,8 @@ def train(self, train):
         batch_size=self.hparams.batch_size,
         epochs=self.hparams.epochs,
         validation_split=0.1,
-        callbacks=callbacks)
+        callbacks=callbacks,
+        verbose=2)
```
ericagreene: This is probably a Keras thing I'm not familiar with, but what does `verbose = 2` mean?

kaggle-classification/keras_trainer/model.py
```diff
@@ -171,15 +173,29 @@ def _setup_embeddings_matrix(self):
...
+  hparams.learning_rate = FLAGS.learning_rate
+  hparams.dropout_rate = FLAGS.dropout_rate
+  hparams.batch_size = FLAGS.batch_size
+
+  trial_log_path = os.path.join(
+      FLAGS.log_path,
+      json.loads(
+          os.environ.get('TF_CONFIG', '{}')
+      ).get('task', {}).get('trial', '')
+  )
```
ericagreene: I don't think this works. 

```
>> export TF_CONFIG=test
>> echo $TF_CONFIG
test
```
```python
import os
import json

json.loads(os.environ.get('TF_CONFIG', '{}'))

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 384, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded
```

Also, I'm just curious why `TF_CONFIG`, `tastk` and `trial` are env variables while all our other parameters are flags.

kaggle-classification/hparam_config.yaml
jjtan: No, but I think we want our python versions to be consistent. I can remove this from this pr though

kaggle-classification/keras_trainer/model.py
jjtan: No, will change back to 1

kaggle-classification/keras_hparam_config.yaml
jjtan: Do the maxTrials / maxParallelTrials above answer that?

kaggle-classification/keras_trainer/model.py
jjtan: Added comment

kaggle-classification/keras_hparam_config.yaml
ericagreene: Got it. Thanks!

PR #49, opened 2018-04-04
Implementation of the Dawid-Skene algorithm 

ericagreene:
This is an application of the Dawid Skene for the toxicity CrowdFlower annotations. It's adapted from an open source implementation of the algorithm [here](https://github.com/dallascard/dawid_skene). 

### background
I read a few papers that proposed using generative probabilistic models to learn better labels from a noisy dataset. Most approaches were adaptations of the [Dawid-Skene algorithm](http://crowdsourcing-class.org/readings/downloads/ml/EM.pdf), which was published in 1979. They propose a simple model of untrustworthy annotators that can be optimized using the EM algorithm to get predictions of the "true" labels and per-annotator error-rates.

This implementation is basically a copy of the approach used in [The Benefits of a Model of Annotation](https://www.aclweb.org/anthology/Q/Q14/Q14-1025.pdf) since they didn't release their code. It takes a few hours to run on a large memory CPU instance for 10 million annotations. 

### to run locally

```
./bin/run_local
```

### to run on ml-engine 
```
./bin/run
```
### todo 
- [x]  Flesh out README
- [x]  Split out E and M code into the fast (matrix multiply) version and the clear (embedded for loop) version. 
- [x] Write out learned rater errors (thetas) to Cloud Storage


annotator-models/README.md
```diff
@@ -0,0 +1,57 @@
...
+    ```
+
+2.  Install library dependencies:
+
+    ```shell
+    pip install -r requirements.txt
+    ```
+
+3.  Create a CSV file of data to train on. The data must be a CSV that has fields:
+    * `_unit_id` corresponding to the item id
```
jjtan: Confirm that this is the correct id field from dataset.

annotator-models/bin/run
```diff
@@ -0,0 +1,63 @@
...
+# Edit these!
+BUCKET_NAME=annotator_models
+CONFIG=cpu_config.yaml
+JOB_NAME=${USER}_kaggle_training
+
+# Note: this must be compatible with cells that have GPUs. us-central1 works.
+# See: https://cloud.google.com/ml-engine/docs/using-gpus
+REGION=us-central1
+DATE=`date '+%Y%m%d_%H%M%S'`
+OUTPUT_PATH=gs://${BUCKET_NAME}/models/${USER}/${DATE}
```
jjtan: this isn't consistent with the other run scripts, we should probably pull this common logic out (but you could just make it consistent for now)

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,441 @@
...
+        log_L = calc_likelihood(counts, class_marginals, error_rates)
+
+        # check for convergence
+        if old_class_marginals is not None:
+            class_marginals_diff = np.sum(np.abs(class_marginals - old_class_marginals))
+            item_class_diff = np.sum(np.abs(item_classes - old_item_classes))
+
+            logging.info('{0}\t{1:.1f}\t{2:.4f}\t{3:.4f}'.format(
+                iter, log_L, class_marginals_diff, item_class_diff))
+
```
jjtan: nit: extra line

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,441 @@
...
+    # initialize
+    iter = 0
+    converged = False
+    old_class_marginals = None
+    old_error_rates = None
+
+    # item_classes is a matrix of estimates of true item classes of size
+    # [items, classes]
+    item_classes = initialize(counts)
+    # item_classes = random_initialization(counts)
```
jjtan: remove

annotator-models/README.md
ericagreene: confirmed :)

ericagreene:
@jjtan This is ready for review whenever you get a chance. I know it's a lot of code (sorry about that!), but I'm happy to walk through it with you. Also, no rush in reviewing.

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,570 @@
...
+
+    # compute class marginals
+    class_marginals = np.sum(item_classes, 0)/float(nItems)
+
+    # compute error rates for each rater, each predicted class
+    # and each true class
+    error_rates = np.matmul(counts.T, item_classes)
+
+    # round very low error rates
+    error_rates[np.abs(error_rates) < 1e-10] = 0
```
jjtan: Comment why

annotator-models/trainer/scratch.txt
```diff
@@ -0,0 +1,43 @@
+e_step, m_step
```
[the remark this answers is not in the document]
ericagreene: Oh no, that's a mistake. Good catch.

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,570 @@
...
+    # while not converged do:
+    while not converged:
+        iter += 1
+
+        # M-step - updated error rates and class marginals given new
+        #          distribution over true item classes
+        old_item_classes = item_classes
+        (class_marginals, error_rates) = m_step(counts, item_classes)
+
+        # E-setp - calculate expected item classes given error rates and
```
nthain: nit: Typo, should be E-step

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,570 @@
...
+    iter = 0
+    converged = False
+    old_class_marginals = None
+    old_error_rates = None
+
+    # item_classes is a matrix of estimates of true item classes of size
+    # [items, classes]
+    item_classes = initialize(counts)
+
+    logging.info('Iter\tlog-likelihood\tdelta-CM\tdelta-Y_hat')
```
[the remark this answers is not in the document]
ericagreene: This is just so the logs print out like:

```
INFO:root:Iter  log-likelihood  delta-CM        delta-Y_hat
INFO:root:1     -4636.3
INFO:root:2     -4005.2         0.1201          216.7878
INFO:root:3     -3702.7         0.0687          175.5724
INFO:root:4     -3681.4         0.0822          120.7572
INFO:root:5     -3604.4         0.0318          59.1577
INFO:root:6     -3559.7         0.0216          40.3534
INFO:root:7     -3554.4         0.0199          27.9144
```

ericagreene:
@jjtan Thanks for reviewing! I think I addressed all your comments. Let me know if there's anything else.

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,565 @@
...
+          [items x classes]
+    """
+    [nItems, nRaters, nClasses] = np.shape(counts)
+
+    item_classes = np.zeros([nItems, nClasses])
+
+    for i in range(nItems):
+
+        counts_i = np.stack([counts[i,:,:],counts[i,:,:]],axis=1)
+        estimate_1 = np.prod(np.power(error_rates,counts_i), axis=(0,2))
```
jjtan: you could maybe move the multiplication by class marginals (i.e. priors) here to more closely mimic eqn 2.5. As in:
 
np.prod(np.power(error_rates,counts_i) * class_marginals, axis=(0,2))

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,565 @@
...
+        item_classes[i,:] = estimate_1
+
+        # normalize error rates by dividing by the sum over all classes
+        item_sum = np.sum(item_classes[i,:])
+        if item_sum > 0:
+            item_classes[i,:] = item_classes[i,:]/float(item_sum)
+
+    return item_classes
+
+def e_step_verbose(counts, class_marginals, error_rates):
```
jjtan: i think we're better off removing the verbose versions and making clearer the vectorized versions

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,565 @@
...
+      item_classes: Soft assignments of items to classes
+          [items x classes]
+    """
+    [nItems, nRaters, nClasses] = np.shape(counts)
+
+    item_classes = np.zeros([nItems, nClasses])
+
+    for i in range(nItems):
+
+        counts_i = np.stack([counts[i,:,:],counts[i,:,:]],axis=1)
```
jjtan: I don't understand why you need to stack two copies

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,565 @@
...
+        estimate_1 = np.prod(np.power(error_rates,counts_i), axis=(0,2))
+        estimate_1 = class_marginals *  estimate_1
+        item_classes[i,:] = estimate_1
```
jjtan: I would leave item_classes alone until the end (i.e. don't use it to hold intermediate values)

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,565 @@
...
+
+        # normalize error rates by dividing by the sum over all classes
```
jjtan: the sum and division can be moved out of the loop

And we should never expect item_sum <= 0 right? Rather than have an if statement it would be better to throw an exception or let it fail (i think)

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,565 @@
...
+    # and each true class
+    error_rates = np.matmul(counts.T, item_classes)
+
+    # reorder axes so its of size [nItems x nClasses x nClasses]
+    error_rates = np.einsum('abc->bca', error_rates)
+
+    # divide each row by the sum of the error rates over all observation classes
+    sum_over_responses = np.sum(error_rates, axis=2)[:,:,None]
+    error_rates = np.divide(error_rates, sum_over_responses,
+                            where=sum_over_responses!=0)
```
jjtan: not sure we need to change this, but am concerned this may hide a bug (since it should never happen)

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,570 @@
...
+    return item_classes
+
+def m_step(counts, item_classes):
+    """
+    Get estimates for the prior class probabilities (p_j) and the error
+    rates (pi_jkl) using MLE with current estimates of true item classes
+    See equations 2.3 and 2.4 in Dawid-Skene (1979)
+
+    Input:
+      counts: Array of how many times each rating was given by each rater
```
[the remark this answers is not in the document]
ericagreene: I'm not using Tensorflow so nothing is a tensor actually :) I think a multi-dimensional array is a pretty common thing to say. But I agree that this comment isn't very clear, it's a good idea to add the dimensions of the array.

annotator-models/trainer/dawid_skene.py
ericagreene: I've found it really useful to have these verbose versions which I know work around when futzing around with the vectorized versions. I can make it clearer that they're not used, maybe add an "appendix" section. There's also 3 different kinds of initialization functions in this file and only 1 is used, but I'd like to keep those too. Does that sound ok? Research code 🤷‍♀️

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,570 @@
...
+    error_rates = np.einsum('abc->bca', error_rates)
+
+    # divide each row by the sum of the error rates over all observation classes
+    sum_over_responses = np.sum(error_rates, axis=2)[:,:,None]
+    error_rates = np.divide(error_rates, sum_over_responses,
+                            where=sum_over_responses!=0)
+
+    return (class_marginals, error_rates)
+
+def m_step_verbose(counts, item_classes):
```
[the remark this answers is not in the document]
ericagreene: See comment below. I will like to keep it but can make it clearer that it's not currently used and rename it.

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,570 @@
...
+    rates (pi_jkl) using MLE with current estimates of true item classes
+    See equations 2.3 and 2.4 in Dawid-Skene (1979)
+
+    Input:
+      counts: Array of how many times each rating was given by each rater
+        for each item
+      item_classes: Matrix of current assignments of items to classes
+
+    Returns:
+      p_j: class marginals [classes]
```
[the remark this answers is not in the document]
ericagreene: Yes

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,517 @@
...
+
+    response_sums = np.sum(counts,1)
+
+    # create an empty array
+    item_classes = np.zeros([nItems, nClasses])
+
+    # for each item, choose a random initial class, weighted in proportion
+    # to the counts from all raters
+    for p in range(nItems):
+        average = response_sums[p,:] / np.sum(response_sums[p,:],dtype=float)
```
jjtan: average -> weights

annotator-models/trainer/dawid_skene.py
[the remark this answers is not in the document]
jjtan: Ah I see. Optional, but you probably want to make each line independently readable so that other log statements can be added without interfering

jjtan:
General comment about the division by zero / NaN checks: Not sure what the right was to do this is, but it seems like in most cases it would be a bug if the zero/NaN occurred - so if we are just absorbing these cases

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,517 @@
...
+        for j in range(nClasses):
+            estimate = class_marginals[j]
+            estimate *= np.prod(np.power(error_rates[:,j,:], counts[i,:,:]))
+
+            item_classes[i,j] = estimate
+
+        # normalize error rates by dividing by the sum over all classes
+        item_sum = np.sum(item_classes[i,:])
+
+        if item_sum == 0 or math.isnan(item_sum):
```
jjtan: General comment about the division by zero / NaN checks: Not sure what the right was to do this is, but it seems like in most cases it would be a bug if the zero/NaN occurred - so if we are ignoring these cases we might not notice a problem. It could be that sometimes it does happen, which is not ideal but fine.. Ideally this never happens and we make these if statements into assertions

annotator-models/trainer/dawid_skene.py
ericagreene: I looked into this and `sum_over_responses` can have values that are 0 if a rater never recorded that

`sum_over_responses` is size `[nRaters, nClasses]`. If rater `k` never labeled any item with class `i`, then `sum_over_responses[k,i]` will be 0. In that case, we want the error rates to be 0, instead of being nan. 

```python
>>> sum_over_responses.shape
(588, 5, 1)

>>> sum_over_responses[587,:,:]
array([[0.5],
       [0.9],
       [0.6],
       [0. ],
       [0. ]])

# recall that error_rates[k, :, :] is a [nClasses, nClasses] array of the error rates for rater k
>>> error_rates[587,:,:]
array([[0.5, 0. , 0. , 0. , 0. ],
       [0.5, 0. , 0.4, 0. , 0. ],
       [0. , 0. , 0.6, 0. , 0. ],
       [0. , 0. , 0. , 0. , 0. ],
       [0. , 0. , 0. , 0. , 0. ]])

# divide error_rates by sum_over_responses, but not when the sum is 0
>>>  np.divide(error_rates, sum_over_responses, where=sum_over_responses!=0)[587,:,:]
array([[1.  , 0.  , 0.  , 0.  , 0.  ],
       [0.56, 0.  , 0.44, 0.  , 0.  ],
       [0.  , 0.  , 1.  , 0.  , 0.  ],
       [0.  , 0.  , 0.  , 0.  , 0.  ],
       [0.  , 0.  , 0.  , 0.  , 0.  ]])

# divide error_rates by sum_over_responses without the condition, see that we get nan
>>> np.divide(error_rates, sum_over_responses)[587,:,:]
array([[1.  , 0.  , 0.  , 0.  , 0.  ],
       [0.56, 0.  , 0.44, 0.  , 0.  ],
       [0.  , 0.  , 1.  , 0.  , 0.  ],
       [ nan,  nan,  nan,  nan,  nan],
       [ nan,  nan,  nan,  nan,  nan]])
```

annotator-models/trainer/dawid_skene.py
```diff
@@ -0,0 +1,565 @@
...
+    for i in range(nItems):
+        item_likelihood = 0.0
+        for j in range(nClasses):
+
+            class_prior = class_marginals[j]
+            item_class_likelihood = np.prod(np.power(error_rates[:,j,:], counts[i,:,:]))
+            item_class_posterior = class_prior * item_class_likelihood
+            item_likelihood += item_class_posterior
+
+        temp = log_L + np.log(item_likelihood)
```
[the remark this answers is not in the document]
ericagreene: This function is written such that the program exits if the likelihood for any example is ever infinity or nan since that should never happen. The equation for the log likelihood function includes an outer sum over all items (the likelihood of the data is the product over the likelihood of each example, the log likelihood is the sum over all examples). We store the likelihood for each example in a `temp` variable, check that it's not infinity or nan and if not, add it to the running sum, ` log_L`.

PR #53, opened 2018-04-17
Speed up vectorized e step 

ericagreene:
I was looking at this code for another reason and had the insight that at each iteration we were doing this computation, which is the same each time since `counts` is the data that doesn't change.

```python
counts_tiled  = np.stack([counts for a in range(nClasses)], axis=2)
```

So I pulled out that computation and each iteration went down from 1.29 seconds to 1 second on 100k annotations.  Now the vectorized `e_step` is at least the same speed as the embedded for-loop version `e_step_verbose`. I'm sure there are more optimizations to be done, but I'm moving on for now.

### DIFF
--- a/annotator-models/trainer/dawid_skene.py
+++ b/annotator-models/trainer/dawid_skene.py
@@ -43,6 +43,11 @@ def run(items, raters, classes, counts, label, tol=1, max_iter=25, init='average
     # item_classes is a matrix of estimates of true item classes of size
     # [items, classes]
     item_classes = initialize(counts)
+    [nItems, nRaters, nClasses] = np.shape(counts)
+
+    # create a tiled version of the counts to speed up calculations in the
+    # e step
+    counts_tiled = np.stack([counts for a in range(nClasses)], axis=2)
 
     logging.info('Iter\tlog-likelihood\tdelta-CM\tdelta-Y_hat\tIter Secs')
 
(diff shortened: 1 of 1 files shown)


annotator-models/trainer/dawid_skene.py
```diff
@@ -204,15 +209,16 @@ def m_step_verbose(counts, item_classes):
...
     Determine the probability of each item belonging to each class,
     given current ML estimates of the parameters from the M-step
     See equation 2.5 in Dawid-Skene (1979)
 
     Inputs:
-      counts: Array of how many times each rating was given
-          by each rater for each item
+      counts_tiled: A matrix of how many times each rating was given
+          by each rater for each item, repeated for each class to make matrix
+          multiplication fasterr. [nItems, nRaters, nClasses, nClasses]
```
jjtan: Can you rename nClasses to distinguish between observation vs truth

ericagreene: Sure. What would you suggest I relabel it to? `nClasses` refers to the number of classes, in the binary case it's 2 and in the toxicity case it's 5 etc.

ericagreene: I updated the doc string. Hopefully it's a bit clearer now, but happy to add more context.

PR #55, opened 2018-04-18
Improve output of the dawid-skene trainer 

ericagreene:
The output of running the dawid-skene trainer is:

* the predicted labels - A file with the per-item predictions, (eg `predictions_obscene_100000.json`) that contains the comment id (`_unit_id`), the mean score from the original data (`obscene_mean`) and predicted probability for each class (`obscene_hat_0`, `obscene_hat_1`).

* the predicted error rates - A file that has the per-rater error rates. Instead of including the full `nClasses x nClasses` error rate matrix, I'm only including the values on the diagonal. I got myself turner around yesterday, because the diagonal error rates are actually accuracy rates (i.e. probability of correctly predicting class k if the true class was k).

This change improves these outputs by 1) updating the column names of the diagonal error rates to per class "accuracy_rates"  and 2) adds the original `comment_text` to the file of predicted labels so we can do analysis on the results without having to join the data in BQ. 

### To test 
Locally, `./bin/run_local`. The output will be written to a local `results` directory.

On ml-engine, `./bin/run`. The output will be written to a Cloud Storage bucket.

### DIFF
--- a/annotator-models/bin/run_local
+++ b/annotator-models/bin/run_local
@@ -7,12 +7,18 @@
 #!/bin/bash
 
 DATE=`date '+%Y%m%d_%H%M%S'`
-BUCKET_NAME=annotator_models
+
+# the path to the CSV with the rating data
+DATA_PATH=gs://annotator_models/dawid_skene_training_data_100k.csv
+
+# the path to the CSV with the full comment text
+COMMENT_TEXT_PATH=gs://annotator_models/q42017_comment_text.csv
 
 gcloud ml-engine local train \
      --module-name=trainer.dawid_skene \
      --package-path=trainer -- \
-     --data-path=gs://${BUCKET_NAME}/dawid_skene_training_data_10k.csv \
-     --label='toxic_score' \
+     --data-path=$DATA_PATH \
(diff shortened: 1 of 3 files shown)


annotator-models/bin/run_local
```diff
@@ -7,12 +7,18 @@
...
+COMMENT_TEXT_PATH=gs://annotator_models/q42017_comment_text.csv
 
 gcloud ml-engine local train \
      --module-name=trainer.dawid_skene \
      --package-path=trainer -- \
-     --data-path=gs://${BUCKET_NAME}/dawid_skene_training_data_10k.csv \
-     --label='toxic_score' \
+     --data-path=$DATA_PATH \
+     --comment-text-path=$COMMENT_TEXT_PATH \
+     --label='obscene' \
```
nthain: Is there a reason to use obscene as the default here? Should we keep it as toxic_score?

ericagreene: The `toxic_score` label has 5 classes and trains much slower unfortunately.  So I'd prefer to keep one of the binary labels there for now. Good question though.

PR #67, opened 2018-05-02
Add smoothing for error rates

ericagreene:
This change adds smoothing to the dawid skene implementation. 

I was noticing that many of the dawid-skene predictions were very close to 0, i.e. for comment i, p(class = j | annotations, worker_error_rates) was often 0. This is the equation for how we calculate this probability given the model: 

screen shot 2018-05-02 at 11 06 17 am

The z's are the predicted labels and the thetas are the worker error rates. If a worker's error rate is zero for one class, which it would be if they never rated any examples with that class, we would never predict that class. It's a standard trick to add pseudo counts to avoid this problem. I've run results with pseudo counts of 3 and saw results improve on hand labeled examples.

### DIFF
--- a/annotator-models/trainer/dawid_skene.py
+++ b/annotator-models/trainer/dawid_skene.py
@@ -23,7 +23,8 @@ import time
 FLAGS = None
 np.set_printoptions(precision=2)
 
-def run(items, raters, classes, counts, label, tol=1, max_iter=25, init='average'):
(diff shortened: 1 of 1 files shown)


ericagreene:
Yep, it's done automatically because we add counts to  the `error_rates` matrix and then sum over that to do the normalization.

PR #69, opened 2018-05-11
Adding attention to GRUs

jjtan:
Performance has not improved yet - but this is the rough architecture I will proceed with.

For each time step we apply a dense layer to the hidden state, then take the results across all time steps and do a dense + sigmoid to get probabilities. These probabilities are then multiplied with the hidden states from before.

### DIFF
--- a/kaggle-classification/keras_trainer/rnn.py
+++ b/kaggle-classification/keras_trainer/rnn.py
@@ -4,7 +4,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-from keras.layers import Input, GRU, Dense, Embedding, Dropout, Bidirectional, TimeDistributed, Multiply, Flatten, Reshape, Dot
+from keras.layers import Input, GRU, Dense, Embedding, Dropout, Bidirectional, TimeDistributed, Flatten, Dot
 from keras.models import Model
 from keras_trainer import base_model
 from keras_trainer.custom_metrics import auc_roc
--- a/kaggle-classification/keras_trainer/rnn.py
+++ b/kaggle-classification/keras_trainer/rnn.py
@@ -4,10 +4,11 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-from keras.layers import Input, GRU, Dense, Embedding, Dropout, Bidirectional, TimeDistributed, Multiply, Flatten, Reshape, GlobalAveragePooling1D
+from keras.layers import Input, GRU, Dense, Embedding, Dropout, Bidirectional, TimeDistributed, Multiply, Flatten, Reshape, Dot
 from keras.models import Model
 from keras_trainer import base_model
 from keras_trainer.custom_metrics import auc_roc
+import keras.backend as K
 
 
 class RNNModel(base_model.BaseModel):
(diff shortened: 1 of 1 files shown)


kaggle-classification/keras_trainer/rnn.py
```diff
@@ -24,15 +24,23 @@ def __init__(self, embeddings_matrix, hparams):
...
   def get_model(self):
-    I = Input(shape=(self.hparams.sequence_length,), dtype='float32')
+    sequence_length = self.hparams.sequence_length
+
+    I = Input(shape=(sequence_length,), dtype='float32')
     E = Embedding(
         self.hparams.vocab_size,
         self.hparams.embedding_dim,
         weights=[self.embeddings_matrix],
         input_length=self.hparams.sequence_length,
```
nthain: Nit: For consistency, I suppose this should be sequence_length instead of self.hparams.sequence_length.

kaggle-classification/keras_trainer/rnn.py
```diff
@@ -24,15 +24,23 @@ def __init__(self, embeddings_matrix, hparams):
...
         trainable=self.hparams.train_embedding)(
             I)
-    X = GRU(128, return_sequences=False)(E)
+    H = Bidirectional(GRU(128, return_sequences=True))(E)
+    A = TimeDistributed(Dense(3), input_shape=(sequence_length, 256))(H)
+    A = Flatten()(A)
+    A = Dense(sequence_length, activation='softmax')(A)
+    A = Reshape((250, 1))(A)
+    X = Multiply()([H, A])
+    X = Flatten()(X)
```
nthain: I'm not sure this is exactly right. Don't you want to sum the weight multiplied hidden units, not flatten them? Otherwise, you're just multiplying by a constant and passing the hidden units along which doesn't feel like it does enough.

jjtan: You're right. Replaced with average pooling... I can't seem to find a way to sum over a dimension. Will try implementing a custom layer later.

jjtan:
Layer (type)                    Output Shape         Param #     Connected to
==================================================================================================
input_1 (InputLayer)            (None, 250)          0
__________________________________________________________________________________________________
embedding_1 (Embedding)         (None, 250, 100)     33925300    input_1[0][0]
__________________________________________________________________________________________________
bidirectional_1 (Bidirectional) (None, 250, 256)     175872      embedding_1[0][0]
__________________________________________________________________________________________________
time_distributed_1 (TimeDistrib (None, 250, 3)       771         bidirectional_1[0][0]
__________________________________________________________________________________________________
flatten_1 (Flatten)             (None, 750)          0           time_distributed_1[0][0]
__________________________________________________________________________________________________
dense_2 (Dense)                 (None, 250)          187750      flatten_1[0][0]
__________________________________________________________________________________________________
dot_1 (Dot)                     (None, 256)          0           bidirectional_1[0][0]
                                                                 dense_2[0][0]
__________________________________________________________________________________________________
dense_3 (Dense)                 (None, 128)          32896       dot_1[0][0]
__________________________________________________________________________________________________
dropout_1 (Dropout)             (None, 128)          0           dense_3[0][0]
__________________________________________________________________________________________________
dense_4 (Dense)                 (None, 6)            774         dropout_1[0][0]
==================================================================================================
Total params: 34,323,363
Trainable params: 398,063
Non-trainable params: 33,925,300

PR #71, opened 2018-05-11
Simplify attention mech in RNN + Fix scripts to handle empty comet_key

jjtan:
using '=' syntax in flags means the empty comet_key is the empty string instead of unset (which causes errors in argparse)

### DIFF
--- a/kaggle-classification/keras_trainer/model.py
+++ b/kaggle-classification/keras_trainer/model.py
@@ -30,7 +30,6 @@ from keras_trainer.single_layer_cnn import SingleLayerCnn
 from keras_trainer.rnn import RNNModel
 from keras_trainer.custom_metrics import auc_roc
 
-
 FLAGS = None
 
 TEMPORARY_MODEL_PATH = 'model.h5'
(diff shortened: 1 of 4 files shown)


kaggle-classification/keras_trainer/rnn.py
```diff
@@ -35,10 +35,11 @@ def get_model(self):
         trainable=self.hparams.train_embedding)(
             I)
     H = Bidirectional(GRU(128, return_sequences=True))(E)
-    A = TimeDistributed(Dense(3), input_shape=(sequence_length, 256))(H)
-    A = Flatten()(A)
-    A = Dense(sequence_length, activation='softmax')(A)
+    A = TimeDistributed(
+        Dense(1, activation='softmax'), input_shape=(sequence_length, 256))(
```
nthain: This looks good. I'd actually add two dense layers (say one of size 128 and then one of size 1). That way the neurons in the output of a single lstm timestep get to interact before deciding how much attention to pay. Does not have to be an immediate fix.

kaggle-classification/keras_trainer/model.py
```diff
@@ -218,13 +224,14 @@ def _setup_embeddings_matrix(self):
   hparams.model_type = FLAGS.model_type
 
   if FLAGS.comet_key:
-    experiment = Experiment(api_key=FLAGS.comet_key, 
-      project_name='comet_trial_run', 
-      auto_param_logging=False,
-      parse_args=False)
+    experiment = Experiment(
```
nthain: Nice. Could you set the default to be the Jigsaw project?

PR #73, opened 2018-05-14
Bugfix: specifying the number of labels needs to change the model head

nthain:
* Changes model architecture in final dense layer depending on number of labels specified
* Fixes score_auc in the case of a single headed model


kaggle-classification/keras_trainer/cnn_with_attention.py
```diff
@@ -35,6 +35,7 @@ class CNNWithAttention(base_model.BaseModel):
   def __init__(self, embeddings_matrix, hparams):
     self.embeddings_matrix = embeddings_matrix
     self.hparams = hparams
+    self.num_labels = len(labels)
```
jjtan: why isn't labels passed in here?

nthain: Good catch! Fixed.

PR #80, opened 2018-05-15
Python type hints [trial]

jjtan:
Initial trial at adding type hints. Not clear if we want to do this yet.

To run the type checker:
mypy --ignore-missing-imports -p keras_trainer


kaggle-classification/keras_trainer/single_layer_cnn.py
```diff
@@ -35,14 +35,15 @@ def __init__(self, embeddings_matrix, hparams, labels):
...
-  def get_model(self):
+  def get_model(self) -> Model:
     I = Input(shape=(self.hparams.sequence_length,), dtype='float32')
     E = Embedding(
         self.hparams.vocab_size,
         self.hparams.embedding_dim,
         weights=[self.embeddings_matrix],
         input_length=self.hparams.sequence_length,
-        trainable=self.hparams.train_embedding)(I)
+        trainable=self.hparams.train_embedding)(
```
iislucas: What's with the changing of spacing/odd line breaks?

PR #82, opened 2018-05-24
Improvements to dawid-skene training 

ericagreene:
This is a bit of an eclectic mix of changes I've made to get Dawid-Skene running on different datasets at scale. 

* Write out prediction and error_rates files as CSVs instead of JSON lines. While JSON lines are a better data format generally, the way Pandas formats them is not compatible with BQ.
* Adds scripts to run on the kaggle dataset for the subtype labels and for toxicity. They are different scripts because the data came from different crowdsourcing jobs. (we should simplify this in the future)
* Reverts to using the verbose (embedded for-loop version) of the e-step because the matrix multiple version runs out of memory on large datasets. 
* Adds a flag to adjust tolerance


annotator-models/bin/run
```diff
@@ -54,15 +55,27 @@ done
...
+	   --config=${CONFIG} \
+	   --module-name=trainer.dawid_skene \
+	   --package-path=trainer \
+	   --region=$REGION \
+	   --verbosity=debug -- \
+	   --data-path=$DATA_PATH \
+	   --comment-text-path=$COMMENT_TEXT_PATH \
+	   --label=$label \
+	   --max-iter=$MAX_ITER \
+           --tolerance=$TOLERANCE \
```
jjtan: format

PR #86, opened 2018-06-05
Estimator framework on ml engine

jjtan:
Also:
  - use tf.app.flags
  - fix dataset input to use iterator (my last model ate all the memory)

### DIFF
--- a/kaggle-classification/tf_trainer/base_keras_model.py
+++ b/kaggle-classification/tf_trainer/base_keras_model.py
@@ -26,7 +26,7 @@ class BaseKerasModel(abc.ABC):
     """
     pass
 
-  def get_estimator(self):
+  def get_estimator(self, model_dir, tmp_model_dir='/tmp/keras_model'):
     """Estimator created based on this instances Keras model.
 
     The generated estimator expected a tokenized text input (i.e. a sequence of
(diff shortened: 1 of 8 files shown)


kaggle-classification/tf_trainer/run.py
```diff
@@ -76,4 +67,4 @@ def main():
 
 
 if __name__ == "__main__":
-  main()
+  tf.app.run(main)
```
nthain: How does this differ from just calling main()?

kaggle-classification/tf_trainer/base_keras_model.py
```diff
@@ -35,14 +35,22 @@ def get_estimator(self):
...
     # does not contain the embedding variable (or other variables that we might
     # want to add outside of the Keras model). The workaround is to specify a
     # model_dir that is *not* the actual model_dir of the final model.
     estimator = tf.keras.estimator.model_to_estimator(
-        keras_model=keras_model, model_dir="/tmp/keras_model")
+        keras_model=keras_model, model_dir=tmp_model_dir)
+
+    new_config = estimator.config.replace(model_dir=model_dir)
+
+    # Why does estimator.model_fn not include params...
```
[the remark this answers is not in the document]
jjtan: Got to look at source for truth: https://github.com/tensorflow/tensorflow/blob/23c218785eac5bfe737eec4f8081fd0ef8e0684d/tensorflow/python/estimator/estimator.py#L126

Config is the optional 5th argument.

My comment was referring to why the property model_fn on an existing estimator is missing the params argument: https://github.com/tensorflow/tensorflow/blob/23c218785eac5bfe737eec4f8081fd0ef8e0684d/tensorflow/python/estimator/estimator.py#L245

kaggle-classification/tf_trainer/run.py
jjtan: got to do it this way to use tf.app.flags (same as Google's absl library)

PR #87, opened 2018-06-08
Add simple test for proto consumption code

jjtan:


kaggle-classification/tf_trainer/tfrecord_input_test.py
```diff
@@ -0,0 +1,49 @@
...
+        labels={"label": tf.float32},
+        word_to_idx={
+            "Hi": 12,
+            "there": 13
+        },
+        unknown_token=999)
+
+    with self.test_session():
+      features, labels = dataset_input._read_tf_example(ex_tensor)
+      self.assertEqual(list(features["comment"].eval()), [12, 13, 999])
```
nthain: Why is there an unknown token at the end of the string "Hi there."?

jjtan: Hypothesis: the period - I bet this tokenizer doesn't remove punctuation but makes it a token and the glove embedding doesn't include it so it is unknown

PR #91, opened 2018-06-15
Model refactor

jjtan:
Sorry not sure how to generate nice diffs (but pretty much duplicated a bunch of files in the original tf_trainer directory with minor corrections for imports). So new directories are keras_gru_attention and tf_gru_attention.

Will probably try to consolidate common run code in a later pr (lots of duplicates in the run scripts).

Other changes:
  - Started using 300 dim embeddings... got tired of crappy aucs.
  - Make common directory.
  - Extract nltk usage from tfrecord_input, and generalize feature preprocessing.


experiments/tf_trainer/tf_gru_attention/run.py
```diff
@@ -0,0 +1,71 @@
...
+LABELS = {
+    "frac_neg": tf.float32,
+    #"frac_very_neg": tf.float32
+}  # type: Dict[str, tf.DType]
+
+
+def main(argv):
+  del argv  # unused
+
+  nltk.download("punkt")
```
nthain: Do we still need this nltk download if we're doing tokenization in the graph? Also the import statement?

jjtan: Yes, still using py_func and relying on nltk to do tokenization. But this refactor allows easy switching.

PR #93, opened 2018-06-18
Common code + Comet ml

jjtan:
Consolidate common code and introduce comet.ml

I wonder if this is a rather opinionated way to structure the code. If it isn't immediately clear whats going on let me know.


experiments/tf_trainer/common/text_preprocessor.py
```diff
@@ -30,7 +30,7 @@ class TextPreprocessor():
   may include fixing a max word length.
   """
 
-  def __init__(self, embeddings_path: types.Path) -> None:
+  def __init__(self, embeddings_path: str) -> None:
```
nthain: Why did we decide to get rid of the Path type? (not that I mind)

experiments/tf_trainer/tf_gru_attention/run.py
```diff
@@ -35,36 +30,52 @@
...
-
-  estimator = preprocessor.create_estimator_with_embedding(
-      estimator_no_embedding, text_feature_name)
+      text_preprocessor=preprocessor)
 
-  for _ in range(20):
-    estimator.train(input_fn=dataset.train_input_fn, steps=1000)
-    metrics = estimator.evaluate(input_fn=dataset.validate_input_fn, steps=100)
-    tf.logging.info(metrics)
+  runner.train_with_eval(20000, 1000, 100)
```
nthain: Because these are serving as templates for future model building, I'd like to make it more obvious what the values that are being given to this function are. Maybe you can make flags (in model runner) with these default values so that everyone knows that these correspond to steps, eval_period, eval_steps.

experiments/tf_trainer/keras_gru_attention/run.py
```diff
@@ -35,36 +30,52 @@
...
-  # TODO: Move embedding into Keras model.
-  estimator = preprocessor.create_estimator_with_embedding(
-      estimator_no_embedding, text_feature_name)
+      text_preprocessor=preprocessor)
 
-  for _ in range(20):
-    estimator.train(input_fn=dataset.train_input_fn, steps=1000)
-    metrics = estimator.evaluate(input_fn=dataset.validate_input_fn, steps=100)
-    tf.logging.info(metrics)
+  runner.train_with_eval(20000, 1000, 100)
```
nthain: Because these are serving as templates for future model building, I'd like to make it more obvious what the values that are being given to this function are. Maybe you can make flags (in model runner) with these default values so that everyone knows that these correspond to steps, eval_period, eval_steps.

experiments/tf_trainer/common/text_preprocessor.py
jjtan: Although I think it's good I just thought the trade off with convenience given that other libraries expected strings for paths made it not worth it.

experiments/tf_trainer/tf_gru_attention/run.py
jjtan: flags + hparams coming up

experiments/tf_trainer/keras_gru_attention/run.py
jjtan: ditto

PR #99, opened 2018-06-25
Script to convert csv to tfrecord

nthain:


experiments/tf_trainer/common/convert_csv_to_tfrecord.py
```diff
@@ -0,0 +1,41 @@
...
+tf.app.flags.DEFINE_string("output_tfrecord_path", None,
+                           "Path where the output TFRecord should be written.")
+
+
+def convert_csv_to_tfrecord(input_csv_path, output_tfrecord_path):
+    df = pd.read_csv(tf.gfile.Open(input_csv_path))
+    csv = df[COLS].values
+    with tf.python_io.TFRecordWriter(output_tfrecord_path) as writer:
+        for row in csv:
+            features, label = row[0], row[1]
```
jjtan: features -> text maybe, since we assume it is text when we encode

experiments/tf_trainer/common/convert_csv_to_tfrecord.py
```diff
@@ -0,0 +1,41 @@
...
+            example = tf.train.Example()
+            example.features.feature[COLS[0]].bytes_list.value.append(features.encode('utf-8'))
```
sorensenjs: when you encode() you might want to add errors='replace'

nthain: Done.

experiments/tf_trainer/common/convert_csv_to_tfrecord.py
nthain: Done.

PR #100, opened 2018-06-27
Fix hparams logging (was wiping hparams when adding the embedding layer)

jjtan:


experiments/tf_trainer/common/base_model.py
```diff
@@ -20,14 +20,16 @@ class BaseModel(abc.ABC):
...
-  def create(
-      estimator_fn: Callable[[str], tf.estimator.Estimator]) -> 'BaseModel':
+  def map(self, f: Callable[[tf.estimator.Estimator], tf.estimator.Estimator]
+         ) -> 'BaseModel':
 
     class Model(BaseModel):
 
-      def estimator(self, model_dir):
-        return estimator_fn(model_dir)
+      def estimator(s, model_dir):
```
nthain: Is s supposed to be self? Or what is s?

experiments/tf_trainer/common/base_model.py
```diff
@@ -20,14 +20,16 @@ class BaseModel(abc.ABC):
...
+        return f(self.estimator(model_dir))
+
+      def hparams(s):
```
nthain: Again should s be self?

experiments/tf_trainer/common/base_model.py
jjtan: renamed it to be clearer

s here is self, however, it refers to the self of the inner class, whereas the self I want to use is from the outer class.

PR #101, opened 2018-06-27
multi task (but labels must exist always)

jjtan:


experiments/tf_trainer/common/model_runner.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, dataset: ds.DatasetInput,
     self._estimator = model.estimator(self._model_dir())
 
   def train_with_eval(self, steps, eval_period, eval_steps):
+    experiment = None
```
nthain: Good catch!

experiments/tf_trainer/keras_gru_attention/run.py
```diff
@@ -36,7 +36,7 @@
 # TODO: Missing fields are not handled properly yet.
 LABELS = {
     "frac_neg": tf.float32,
-    #"frac_very_neg": tf.float32
+    "frac_very_neg": tf.float32
```
nthain: Is there a plan to make the labels a flag?

experiments/tf_trainer/tf_gru_attention/model.py
```diff
@@ -31,7 +32,7 @@
 
 class TFRNNModel(base_model.BaseModel):
 
-  def __init__(self, text_feature_name, target_labels):
+  def __init__(self, text_feature_name: str, target_labels: List[str]) -> None:
```
nthain: Here you suggest the type of target_labels should be a list, but in run.py you call it with a dict.

experiments/tf_trainer/keras_gru_attention/run.py
jjtan: Maybe. I think I will first experiment a bit with the idea we discussed where I associate the heads (and labels with them) to the dataset_input. My thought is that the labels are inherently linked to a dataset

experiments/tf_trainer/tf_gru_attention/model.py
jjtan: Let me fix that... mypy is failing me

PR #102, opened 2018-06-28
Adding Keras CNN model

nthain:


experiments/tf_trainer/keras_cnn/model.py
```diff
@@ -0,0 +1,92 @@
...
+        name='comment_text')
+
+    # Convolutional Layers
+    X = I
+    for filter_size in self.hparams().filter_sizes:
+        X = layers.Conv1D(self.hparams().num_filters, filter_size, activation='relu', padding='same')(X)
+
+
+    X = layers.GlobalAveragePooling1D()(X)
+    #X = layers.Flatten()(X)
```
jjtan: remove?

experiments/tf_trainer/keras_cnn/model.py
```diff
@@ -0,0 +1,92 @@
...
+    # Outputs
+    outputs = []
+    for label in self._labels:
+      outputs.append(layers.Dense(1, activation='sigmoid', name=label)(X))
+
+    model = models.Model(inputs=I, outputs=outputs)
+    model.compile(
+        optimizer=optimizers.Adam(lr=self.hparams().learning_rate),
+        loss='binary_crossentropy',
+        metrics=['accuracy', super().roc_auc])
```
jjtan: this might be where my bug is... I don't think you should wait, but you may have to change this later to support multihead

experiments/tf_trainer/keras_cnn/run_ml_engine.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
jjtan: Random note: seeing such a similar file is motivating me to figure out some way to make this common

experiments/tf_trainer/keras_cnn/model.py
nthain: Done.

jjtan:
Maybe you aren't running into the issue then.. What I was expecting was really bad roc_aucs. Accuracy might be a separate problem / might be a metric problem (you can try 'binary_accuracy' maybe)

PR #109, opened 2018-06-29
tweaks to comments and docstrings

iislucas:


experiments/tf_trainer/common/model_runner.py
```diff
@@ -51,7 +51,14 @@ def __init__(self, dataset: ds.DatasetInput,
     self._model = model
     self._estimator = model.estimator(self._model_dir())
 
+  # TODO(ldixon): consider early stopping. Currently steps is  hard coded.
```
nthain: Nit: Looks like there is an extra space between 'is' and 'hard' coded.

experiments/tf_trainer/common/base_model.py
```diff
@@ -22,6 +22,9 @@ class BaseModel(abc.ABC):
 
   def map(self, f: Callable[[tf.estimator.Estimator], tf.estimator.Estimator]
          ) -> 'BaseModel':
+    """
+    Allows models to be extended. e.g. adding preprocessing steps.
```
jjtan: i think to be in line with Google python style you want the first sentence to be directly after the first quotes:
"""Comment here"""

Or maybe:
"""Allows models to be extended.

For example, adding preprocessing steps
"""

experiments/tf_trainer/common/model_runner.py
```diff
@@ -51,7 +51,14 @@ def __init__(self, dataset: ds.DatasetInput,
...
   def train_with_eval(self, steps, eval_period, eval_steps):
+    """
```
jjtan: Same as above, I think the style guide requires some text after the initial quotes. Maybe:

"""Trains model.

Args:
...
"""

experiments/tf_trainer/keras_cnn/model.py
```diff
@@ -22,6 +22,7 @@
                           'The dropout rate to use during training.')
 # This would normally just be a multi_integer, but we use string due to
 # constraints with ML Engine hyperparameter tuning.
+# TODO: add link to relevant public issue/bug/documentation?
```
jjtan: https://cloud.google.com/ml-engine/docs/tensorflow/hyperparameter-tuning-overview#hyperparameter_types

PR #112, opened 2018-06-29
Revert multiheadedness

jjtan:


jjtan:
Sorry I had meant to write more. Bug is not fixed, but I am already seeing > 0.7 AUCs again for Keras (whereas before it hovered around 0.5). This is a just to have master not broken, am still investigating.

PR #115, opened 2018-06-29
Tf bidirectional GRU attention [DO NOT MERGE]

jjtan:


experiments/tf_trainer/tf_gru_attention/model.py
```diff
@@ -60,25 +60,31 @@ def _model_fn(self, features, labels, mode, params, config):
...
         dtype=tf.float32)
 
     # TF needs help understanding sequence length (I think because we're using
     # dynamic_rnn)
-    outputs = tf.reshape(
-        outputs, [batch_size, params.max_seq_length, params.gru_units[-1]])
+    # TODO: We're using outputs and not states... is this okay?
+    # TODO: Convert to using static_state_saving_rnn and use states
+    last_gru_units = params.gru_units[-1] * 2  # x2 because bidirectional
+    outputs = tf.reshape(outputs,
```
nthain: I'm a little confused. Isn't 'outputs' a tuple of a forward and backward tensor? Can you use reshape on a tuple in this way?

jjtan: You're right, I'm not sure what's going and will have to run some more experiments later but this performs better than after I concat the two...

PR #121, opened 2018-07-02
Move to using .pyenv (standard) for venv; and move convert to tools dir

iislucas:


iislucas:
(Oops, I just noticed I merged in a different pull request that was supposed to go to master... luckily it was only 3 lines, but what that means is that this PR now has also removed some unused deps.)

experiments/README.md
```diff
@@ -20,11 +20,17 @@ Install library dependencies (it is optional, but recommended to install these
...
+    # The python3 way to create and use virtual environment
+    # (optional, but recommended):
+    python3 -m venv .pyenv
+    source .pyenv/bin/activate
     # Install dependencies
     pip install -r requirements.txt
+
+    # ... do stuff ...
+
+    # Exist your virtual environment.
```
nthain: Exist -> Exit

iislucas: done

PR #123, opened 2018-07-02
Use Bazel; specify tests; update README

iislucas:
Uses Bazel to enable running all of the tests hermetically, quickly, and other nice build things.


iislucas:
Oops, this should be rebased on aimed at master; and land after iislucas-3-fix-deps lands. Closing for now.

PR #126, opened 2018-07-10
Export model

fprost:
Please review my changes and suggest any improvements!

### DIFF
--- a/experiments/tf_trainer/common/serving_input.py
+++ b/experiments/tf_trainer/common/serving_input.py
@@ -1,0 +1,28 @@
+'''Serving functions for deployed model.'''
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow as tf
+from tensorflow.python.ops import array_ops
+
+def create_serving_input_fn(feature_preprocessor_init, sentence_name, key_name):
+
+  def serving_input_fn():
+    features_placeholders = {}
+    features_placeholders[sentence_name] = array_ops.placeholder(
+        dtype=tf.string, name=sentence_name)
+    features_placeholders[key_name] = array_ops.placeholder(
+        dtype=tf.string, name=key_name)
+
+    features = features_placeholders
+    feature_preprocessor = feature_preprocessor_init()
+    features[sentence_name] = feature_preprocessor(
+      features_placeholders[sentence_name]) 
+
+    return tf.estimator.export.ServingInputReceiver(
+        features,
+        features_placeholders)
+
+  return serving_input_fn
--- a/experiments/tf_trainer/common/serving_input.py
+++ b/experiments/tf_trainer/common/serving_input.py
@@ -16,10 +16,11 @@ def create_serving_input_fn(feature_preprocessor_init, text_feature_name, key_na
     features_placeholders[key_name] = array_ops.placeholder(
         dtype=tf.string, name=key_name)
 
-    features = features_placeholders
+    features = {}
+    features[key_name] = features_placeholders[key_name]
     feature_preprocessor = feature_preprocessor_init()
     features[text_feature_name] = feature_preprocessor(
-      features_placeholders[text_feature_name]) 
+      features_placeholders[text_feature_name])
 
     return tf.estimator.export.ServingInputReceiver(
         features,
--- a/experiments/tf_trainer/common/serving_input.py
+++ b/experiments/tf_trainer/common/serving_input.py
@@ -7,19 +7,19 @@ from __future__ import print_function
 import tensorflow as tf
 from tensorflow.python.ops import array_ops
 
-def create_serving_input_fn(feature_preprocessor_init, sentence_name, key_name):
+def create_serving_input_fn(feature_preprocessor_init, text_feature_name, key_name):
 
   def serving_input_fn():
     features_placeholders = {}
-    features_placeholders[sentence_name] = array_ops.placeholder(
-        dtype=tf.string, name=sentence_name)
+    features_placeholders[text_feature_name] = array_ops.placeholder(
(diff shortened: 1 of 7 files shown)


fprost:
@nthain, it seems I needed to add you while creating the pull request. Let me know if there is a way to fix that.
Thanks

experiments/tf_trainer/common/tfrecord_input.py
```diff
@@ -60,10 +63,14 @@ def _input_fn_from_file(self, filepath: str) -> types.FeatureAndLabelTensors:
     # TODO: think about what happens when we run out of examples; should we be
     # using something that repeats over the dataset many time to allow
     # multi-epoch learning, or does estimator do this for us?
-    itrOp = batched_dataset.make_one_shot_iterator().get_next()
-    return itrOp
-
-  def _read_tf_example(self, record: tf.Tensor) -> types.FeatureAndLabelTensors:
+    itr_op = batched_dataset.make_initializable_iterator()
+    tf.add_to_collection(tf.GraphKeys.TABLE_INITIALIZERS, itr_op.initializer)
```
jjtan: Could you add a comment explaining what this is doing?

experiments/tf_trainer/keras_gru_attention/run.py
```diff
@@ -45,17 +51,21 @@ def main(argv):
 
   embeddings_path = FLAGS.embeddings_path
   text_feature_name = FLAGS.text_feature_name
+  key_name = FLAGS.key_name
 
   preprocessor = text_preprocessor.TextPreprocessor(embeddings_path)
-  nltk.download("punkt")
-  tokenize_op = preprocessor.tokenize_tensor_op(nltk.word_tokenize)
+  if FLAGS.preprocess_in_tf:
+    tokenize_op_init = lambda: preprocessor.tokenize_tensor_op_tf_func()
```
jjtan: Can definitely be later work, but we could maybe unify the tf_func and py_func version of this by passing in a tf.py_func in the case we want to use a python function.

experiments/tf_trainer/common/text_preprocessor.py
```diff
@@ -32,13 +30,57 @@ class TextPreprocessor():
...
+
+      Args:
+        text: must be a scalar string tensor (rank 0).
+
+      Returns:
+        A 1-D Tensor of word integers.
+      '''
+
+      # TODO: Improve tokenizer.
+      # TODO: Add utf-8 encoding.
```
nthain: Could you expand on this TODO a bit? How does this function currently handle utf-8 text?

experiments/tf_trainer/keras_gru_attention/run.deploy.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# Edit these!
+MODEL_NAME=keras_gru_attention
+MODEL_SAVED_PATH=gs://kaggle-model-experiments/${USER}/1531159115
```
nthain: Is the hardcoded number in the saved path intentional? Is it better to replace with a timestamp?

experiments/tf_trainer/common/serving_input.py
```diff
@@ -0,0 +1,28 @@
...
+def create_serving_input_fn(feature_preprocessor_init, sentence_name, key_name):
+
+  def serving_input_fn():
+    features_placeholders = {}
+    features_placeholders[sentence_name] = array_ops.placeholder(
+        dtype=tf.string, name=sentence_name)
+    features_placeholders[key_name] = array_ops.placeholder(
+        dtype=tf.string, name=key_name)
+
+    features = features_placeholders
```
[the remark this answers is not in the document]
fprost: Looking at tf.estimator.export.ServingInputReceiver: 
1st argument = transformed features
2nd argument: raw features (placeholders).

In our case, both have the same field names but we apply some preprocessing on the 'sentence_name' field

experiments/tf_trainer/common/serving_input.py
```diff
@@ -0,0 +1,28 @@
+'''Serving functions for deployed model.'''
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow as tf
+from tensorflow.python.ops import array_ops
+
+def create_serving_input_fn(feature_preprocessor_init, sentence_name, key_name):
```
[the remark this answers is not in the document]
fprost: DONE

experiments/tf_trainer/common/serving_input.py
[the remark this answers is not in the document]
nthain: But since we're aliasing, aren't we applying the preprocessing to the sentence_name field of both features and features_placeholders?

experiments/tf_trainer/keras_gru_attention/run.deploy.sh
fprost: Good question. This path needs to be the same as the model_dir in run.cloud.sh.
I am not sure there is an automatic way to extract it.

If we aggree that the path will always follow that pattern "gs://kaggle-model-experiments/${USER}/$TIMESTAMP" then for deployment we can take (as default) the latest timestamp.

What do you think?

experiments/tf_trainer/common/text_preprocessor.py
fprost: Added some info.

experiments/tf_trainer/keras_gru_attention/run.deploy.sh
nthain: That makes sense. Though, the path we've been defaulting to has been:
"gs://kaggle-model-experiments/tf_trainer_runs/${USER}/${MODEL_NAME}/${DATETIME}/model_dir". What do you think?

PR #131, opened 2018-07-16
Typo fix

deeglaze:
I would have suggested "out" to be "our" as the most readily available fix, but open source ownership trickier than that. You could state it as "our" and sound closed to contributes, "your" and make the project sound like an abandoned code dump, or "the result of core team and community contributions", which is both verbose and makes an "us vs them" comparison for contributors, etc. Thus I chose to describe the model as a component of the project rather than a component owned by some group of people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ This repository is contains example code to train machine learning models for te
 
 # Outline of the codebase
 
-* `experiments/` contains the conversation AI ML training framework.
+* `experiments/` contains the ML training framework.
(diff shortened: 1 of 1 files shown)


nthain:
Thanks for the catch and the fix. I made a small adjustment as I don't want this to be seen as "the" conversation-ai ML framework as conversation-ai has several and that could lead to confusion. Having said that, your point is well made and I like the new description much more than the old one.

PR #137, opened 2018-07-17
Adding evaluation pipeline only

fprost:
Thanks for your review!


fprost:
Edits made. 
Will create issue for "make batch predictions into it's own library".

PR #140, opened 2018-07-24
Change embedding

fprost:
This commit includes:
- A flag to say if embedding file is 'rb' or 'r'. I tried to extract automatically this from the file but it was trickier than expected: some embeddings can be read 50% with 'r' before throwing an error. That means that, if you want to check the file format and start looking at the wrong one, you will go 1.5 through the embedding size. --> this is very slow. I prefer to trust the user and putting an error message to highlight the parameter to the user.
As default is 'r', this should be compatible with other branches as Glove is 'r'.

- The fix to handle large embeddings.
This should also be compatible with other branches.

Let me know what you think!

### DIFF
--- a/experiments/tf_trainer/common/text_preprocessor.py
+++ b/experiments/tf_trainer/common/text_preprocessor.py
@@ -29,10 +29,14 @@ class TextPreprocessor(object):
   may include fixing a max word length.
   """
 
-  def __init__(self, embeddings_path: str) -> None:
+  def __init__(self, 
+               embeddings_path: str, 
+               is_binary_embedding: Optional[bool] = False
+               ) -> None:
     self._word_to_idx, self._embeddings_matrix, self._unknown_token = \
         TextPreprocessor._get_word_idx_and_embeddings(
-            embeddings_path)  # type: Tuple[Dict[str, int], np.ndarray, int]
+            embeddings_path,
+            is_binary_embedding)  # type: Tuple[Dict[str, int], np.ndarray, int]
 
   def tokenize_tensor_op_tf_func(self) -> Callable[[types.Tensor], types.Tensor]:
     """Tensor op that converts some text into an array of ints that correspond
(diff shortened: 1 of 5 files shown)


experiments/tf_trainer/common/text_preprocessor.py
```diff
@@ -193,12 +236,19 @@ def _get_word_idx_and_embeddings(embeddings_path: str,
...
+    else:
+      read_mode = 'r'
+    with tf.gfile.Open(embeddings_path, read_mode) as f:
       for idx, line in enumerate(f):
         if max_words and idx >= max_words:
           break
 
         values = line.split()
+        # Remove header when necessary.
+        if len(values) == 2 and idx == 0:
```
nthain: Is it better to have this be automated or as a flag? My guess is that it's better to have an explicit flag than this check which could fail implicitly for lots of reasons.

experiments/tf_trainer/keras_gru_attention/run.ml_engine.sh
```diff
@@ -22,7 +22,7 @@ gcloud ml-engine jobs submit training tf_trainer_${MODEL_NAME}_${USER}_${DATETIM
     -- \
     --train_path="${GCS_RESOURCES}/toxicity_q42017_train.tfrecord" \
     --validate_path="${GCS_RESOURCES}/toxicity_q42017_validate.tfrecord" \
-    --embeddings_path="${GCS_RESOURCES}/glove.6B/glove.6B.300d.txt" \
+    --embeddings_path="${GCS_RESOURCES}/google-news/GoogleNews-vectors-negative300.txt" \
```
nthain: Is this the debiased word embedding?

experiments/tf_trainer/common/text_preprocessor.py
```diff
@@ -116,15 +120,46 @@ def create_estimator_with_embedding(
...
+    Note: We need to consider the case of large embeddings (see: 
+      https://stackoverflow.com/questions/48217599/
+      how-to-initialize-embeddings-layer-within-estimator-api/48243086#48243086).
+
     """
     old_model_fn = estimator.model_fn
     old_config = estimator.config
     old_params = estimator.params
 
+    def add_init_fn_to_estimatorSpec(estimator_spec, init_fn):
```
nthain: Thanks for writing this, but wow it seems complex. Why is this necessary?

experiments/tf_trainer/keras_gru_attention/run.ml_engine.sh
fprost: This is the unbiased one.

experiments/tf_trainer/common/text_preprocessor.py
fprost: From our discussion, should I add information in here?

experiments/tf_trainer/common/text_preprocessor.py
fprost: Hmm good point.
So the solution should be:  flag "skip_first_line"?

Will add it to next commit.

PR #143, opened 2018-07-25
Update tensorflow branch

fprost:
Contains:
- Adaptation of run.py for tf model
- Change of attention layers
- A new shell script


experiments/tf_trainer/tf_gru_attention/run.py
```diff
@@ -22,15 +22,22 @@
 tf.app.flags.DEFINE_string("embeddings_path",
                            "local_data/glove.6B/glove.6B.100d.txt",
                            "Path to the embeddings file.")
+tf.app.flags.DEFINE_boolean("is_binary_embedding", True,
```
nthain: We should probably have the default flag match the default embedding path. So, that would make it False, right?

experiments/tf_trainer/tf_gru_attention/run.py
```diff
@@ -44,27 +51,40 @@ def main(argv):
...
   dataset = tfrecord_input.TFRecordInput(
       train_path=FLAGS.train_path,
       validate_path=FLAGS.validate_path,
       text_feature=text_feature_name,
       labels=LABELS,
-      feature_preprocessor=tokenize_op,
+      feature_preprocessor_init=tokenize_op_init,
       batch_size=FLAGS.batch_size)
 
+  # TODO: Move embedding *into* Keras model.
```
nthain: What does this mean?

experiments/tf_trainer/tf_gru_attention/run.py
fprost: Yes good point. Will change

experiments/tf_trainer/tf_gru_attention/run.py
fprost: This comment was written by someone else.

I am assuming that the goal here would be to have the embeddings directly in the model function (keras and tf model.py) and therefore get rid of this function add_embedding_to_model.

PR #145, opened 2018-07-26
Dynamic padding clean

fprost:
I implemented here the dynamic padding (with the bucket_by_sequence_length). One final improvement could be to reject large sentences.
The model files have been changed to be resistant to variable sequence_length.

While I was working on it, I realized there was a bug in the nltk tokenizer (surprising since I had tested it but I must have messed it up afterwards). Here is the solution as well.


experiments/tf_trainer/common/text_preprocessor.py
```diff
@@ -172,8 +186,8 @@ def new_model_fn(features, labels, mode, params, config):
       text_feature = features[text_feature_name]
       # Make sure all examples are length 300
       # TODO: Parameterize 300
-      text_feature = tf.pad(text_feature, [[0, 0], [0, 300]])
-      text_feature = text_feature[:, 0:300]
+      # text_feature = tf.pad(text_feature, [[0, 0], [0, 300]])
```
nthain: Let's remove these commented lines. They will still exist in the revision history if needed.

fprost: Missed that! Thanks and done!

PR #146, opened 2018-08-13
Count unknown words

fprost:
Nithum,

The goal here is to count the number of unknown words. What I do not like in my code (besides added complexity) is the hardcoded 100   (should be set to the default of the summary writer but not sure how to access it.

Not sure if this PR is worth merging, but creating it at least for potential future references.

### DIFF
--- a/experiments/Analyze preprocessing.ipynb
+++ b/experiments/Analyze preprocessing.ipynb
@@ -1,0 +1,525 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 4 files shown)


fprost:
I struggled to add it as a tf.summary and am currently printing it to the screen.

nthain:
Let's discuss tomorrow in our meeting to determine if this should be merged.

PR #149, opened 2018-08-28
Serving for tf

fprost:
Adding the serving function for tf-part.

2 points:
- In model_trainer, I had to rewrite the forward_features function to workaround a bug. I created an internal bug to update the function.
- is_trainable is void for a keras model. I can explain why in our meeting but it might require some work to get it fixed. Maybe we should raise an exception when trainable=True and model=keras? Not sure where to insert that though.

### DIFF
--- a/experiments/tf_trainer/common/model_trainer.py
+++ b/experiments/tf_trainer/common/model_trainer.py
@@ -11,10 +11,20 @@ from __future__ import print_function
 import json
 import os
 import os.path
+import six
 
 import comet_ml
 import tensorflow as tf
 from tensorflow.python.platform import tf_logging as logging
+from tensorflow.python.estimator import estimator as estimator_lib
+from tensorflow.python.estimator import model_fn as model_fn_lib
+from tensorflow.python.estimator.export.export_output import PredictOutput
+from tensorflow.python.framework import ops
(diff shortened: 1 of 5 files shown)


experiments/tf_trainer/common/model_trainer.py
```diff
@@ -41,32 +51,120 @@
...
+        raise ValueError(
+            'Feature ({}) should be a Tensor. Please use `keys` '
+            'argument of forward_features to filter unwanted features, or'
+            'add key to argument `sparse_default_values`.'
+            'Type of features[{}] is {}.'.format(key, key, type(feature)))
+      predictions[key] = feature
+    spec = spec._replace(predictions=predictions)
+    if spec.export_outputs: # CHANGES HERE
+      outputs = spec.export_outputs['predict'].outputs
+      outputs['comment_key'] = spec.predictions['comment_key']
```
nthain: Where do the keys like 'predict' and 'comment_key' come from? Are they hardcoded for our purposes or some kind of general keys used by EstimatorSpec?

fprost: Good catch. comment_key was definitely an error from my end.

for 'predict' and 'serving_default' these are general keys by estimatorSpec. Let me know if I can leave them as is

PR #150, opened 2018-09-05
Train and evaluate

fprost:
This commit aims at correcting this bug: https://github.com/tensorflow/tensorflow/issues/19062.
The way to make the correction is to upgrade to tf==1.10 and to use train_and_evaluate.

Using train_and_evaluate removes the flexibility around the training loop. In particular, it is not easy anymore to include comet. We have removed it for now and could potentially be re-introduced via training_hooks.

### DIFF
--- a/experiments/tf_trainer/common/tfrecord_input.py
+++ b/experiments/tf_trainer/common/tfrecord_input.py
@@ -57,7 +57,7 @@ class TFRecordInput(dataset_input.DatasetInput):
             {self._text_feature: [None],
             'sequence_length': []},
             {label: [] for label in self._labels})
-    # TODO(fprost_: Remove long sentences.
+    # TODO(fprost): Set a filter to remove long sentences (for training).
     parsed_dataset = parsed_dataset.apply(
         tf.contrib.data.bucket_by_sequence_length(
             element_length_func=lambda x,_: x['sequence_length'],
--- a/experiments/tf_trainer/common/tfrecord_input.py
+++ b/experiments/tf_trainer/common/tfrecord_input.py
@@ -27,7 +27,7 @@ class TFRecordInput(dataset_input.DatasetInput):
       train_preprocess_fn: Callable[[str], List[str]],
       batch_size: int = 64,
       round_labels: bool = True,
-      num_prefetch: int = 3) -> None:
+      num_prefetch: int = 5) -> None:
     self._train_path = train_path
     self._validate_path = validate_path
     self._text_feature = text_feature
@@ -46,6 +46,7 @@ class TFRecordInput(dataset_input.DatasetInput):
     return self._input_fn_from_file(self._validate_path)
 
   def _input_fn_from_file(self, filepath: str) -> types.FeatureAndLabelTensors:
+
     dataset = tf.data.TFRecordDataset(filepath)  # type: tf.data.TFRecordDataset
 
     parsed_dataset = dataset.map(
@@ -56,21 +57,16 @@ class TFRecordInput(dataset_input.DatasetInput):
             {self._text_feature: [None],
             'sequence_length': []},
             {label: [] for label in self._labels})
-    # TODO: Remove long sentences.
+    # TODO(fprost_: Remove long sentences.
     parsed_dataset = parsed_dataset.apply(
         tf.contrib.data.bucket_by_sequence_length(
             element_length_func=lambda x,_: x['sequence_length'],
             bucket_boundaries=[(i + 1) * 20 for i in range(10)],
             bucket_batch_sizes=[self._batch_size] * 11,
             padded_shapes=padded_shapes)
-        )
+        )        
     batched_dataset = parsed_dataset.prefetch(self._num_prefetch)
-
-    itr_op = batched_dataset.make_initializable_iterator()
-    # Adding the initializer operation to the graph.
-    tf.add_to_collection(tf.GraphKeys.TABLE_INITIALIZERS, itr_op.initializer)
-    
-    return itr_op.get_next()
+    return batched_dataset
 
   def _read_tf_example(self, record: tf.Tensor,
                       ) -> types.FeatureAndLabelTensors:
--- a/experiments/requirements.txt
+++ b/experiments/requirements.txt
@@ -22,7 +22,7 @@ requests==2.19.1
 scipy==1.1.0
 six==1.11.0
 tensorboard==1.8.0
-tensorflow==1.8.0
+tensorflow==1.10.0
 termcolor==1.1.0
 typed-ast==1.1.0
 urllib3==1.23
--- a/experiments/tf_trainer/common/model_trainer.py
+++ b/experiments/tf_trainer/common/model_trainer.py
@@ -185,41 +185,27 @@ class ModelTrainer(object):
       eval_period: the number of steps between evaluations.
       eval_steps: the number of batches that are evaluated per evaulation.
     """
-    experiment = None
-    if FLAGS.comet_key_file is not None:
-      experiment = self._setup_comet()
-    num_itr = int(steps / eval_period)
 
-    for _ in range(num_itr):
-      hooks = None
-      if FLAGS.enable_profiling:
-        hooks = [tf.train.ProfilerHook(save_steps=10,
-                                       output_dir=os.path.join(self._model_dir(), 'profiler'))]
-      self._estimator.train(
-          input_fn=self._dataset.train_input_fn,
-          steps=eval_period,
-          hooks=hooks)
-      metrics = self._estimator.evaluate(
-          input_fn=self._dataset.validate_input_fn, steps=eval_steps)
-      if experiment is not None:
-        tf.logging.info('Logging metrics to comet.ml: {}'.format(metrics))
-        experiment.log_multiple_metrics(metrics)
-      tf.logging.info(metrics)
+    training_hooks = None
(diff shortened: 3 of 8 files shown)


nthain:
@jjtan I'll review this, I'm just adding you as an assignee to cc you on this. I know that you'll be building on this codebase and just wanted to make you aware of this change.

experiments/tf_trainer/common/tfrecord_input.py
```diff
@@ -56,21 +57,16 @@ def _input_fn_from_file(self, filepath: str) -> types.FeatureAndLabelTensors:
             {self._text_feature: [None],
             'sequence_length': []},
             {label: [] for label in self._labels})
-    # TODO: Remove long sentences.
+    # TODO(fprost_: Remove long sentences.
```
nthain: Nit: Typo? (should it be '(fprost)')

Also, could you expand on what this means?

fprost:
Thanks updated comment.

fprost:
For more information:
We noticed that the model was quickly overfitting (before even going through 1/10 of the training data) and tracked down to a bug from the training loop.

As this TF [issue](https://github.com/tensorflow/tensorflow/issues/19062) explains, when we use the Dataset API and call successively `train` and `eval`, the states of the queues are not preserved. In other words, every time we call `train`, we read the training data from the beginning! 
TF team has fixed it (partially) in tf>=1.9 for `train_and_evaluate`, which keeps the state of the training queue (and not eval).

For that reason, we decided to use `train_and_evaluate` and that solved this bad behavior.
However, because you have no control of the training loop with this method, we could not export the comet files as before. So we lost the comet feature in this commit.

We brainstormed about how to keep comet while fixing the bug. Potentially, we could have:
1- Keep our training loop and get inspired by `train_and_evaluate` to write code to keep the queue states. We were reluctant to do that because it seemed relatively complicated and hard to maintain.
2- Use `train_and_evaluate` and exporting the comet files with training hooks. I believe this option is relatively coherent, however we did not have time to pursue.

iislucas:
So I jumped onto Slack and asked comel ml people who were super helpful and quick - turns out they automatically report everything that’s reported to Tensorboard. So generally speaking @nthain  could have avoided the headache of pulling the tensors from inside the estimator in PR#150 by just having an experiment object somewhere. So should actually be easy to use comet even with estimator framework. 

Code sample: 
```
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function
from comet_ml import Experiment
import os
import time
import numpy as np
import tensorflow as tf
import six
from tensorflow.python.framework import ops
from tensorflow.python.platform import tf_logging as logging
from tensorflow.python.training import session_run_hook
from tensorflow.python.training import training_util
from tensorflow.python.training.session_run_hook import SessionRunArgs
from tensorflow.python.training.summary_io import SummaryWriterCache
from tensorflow.python.util.tf_export import tf_export
class _HookTimer(object):
    """Base timer for determining when Hooks should trigger.
    Should not be instantiated directly.
    """
    def __init__(self):
        pass
    def reset(self):
        """Resets the timer."""
        pass
    def should_trigger_for_step(self, step):
        """Return true if the timer should trigger for the specified step."""
        raise NotImplementedError
    def update_last_triggered_step(self, step):
        """Update the last triggered time and step number.
        Args:
          step: The current step.
        Returns:
          A pair `(elapsed_time, elapsed_steps)`, where `elapsed_time` is the number
          of seconds between the current trigger and the last one (a float), and
          `elapsed_steps` is the number of steps between the current trigger and
          the last one. Both values will be set to `None` on the first trigger.
        """
        raise NotImplementedError
    def last_triggered_step(self):
        """Returns the last triggered time step or None if never triggered."""
        raise NotImplementedError
@tf_export("train.SecondOrStepTimer")
class SecondOrStepTimer(_HookTimer):
    """Timer that triggers at most once every N seconds or once every N steps.
    """
    def __init__(self, every_secs=None, every_steps=None):
        self.reset()
        self._every_secs = every_secs
        self._every_steps = every_steps
        if self._every_secs is None and self._every_steps is None:
            raise ValueError(
                "Either every_secs or every_steps should be provided.")
        if (self._every_secs is not None) and (self._every_steps is not None):
            raise ValueError(
                "Can not provide both every_secs and every_steps.")
        super(SecondOrStepTimer, self).__init__()
    def reset(self):
        self._last_triggered_step = None
        self._last_triggered_time = None
    def should_trigger_for_step(self, step):
        """Return true if the timer should trigger for the specified step.
        Args:
          step: Training step to trigger on.
        Returns:
          True if the difference between the current time and the time of the last
          trigger exceeds `every_secs`, or if the difference between the current
          step and the last triggered step exceeds `every_steps`. False otherwise.
        """
        if self._last_triggered_step is None:
            return True
        if self._last_triggered_step == step:
            return False
        if self._every_secs is not None:
            if time.time() >= self._last_triggered_time + self._every_secs:
                return True
        if self._every_steps is not None:
            if step >= self._last_triggered_step + self._every_steps:
                return True
        return False
    def update_last_triggered_step(self, step):
        current_time = time.time()
        if self._last_triggered_time is None:
            elapsed_secs = None
            elapsed_steps = None
        else:
            elapsed_secs = current_time - self._last_triggered_time
            elapsed_steps = step - self._last_triggered_step
        self._last_triggered_time = current_time
        self._last_triggered_step = step
        return (elapsed_secs, elapsed_steps)
    def last_triggered_step(self):
        return self._last_triggered_step
class NeverTriggerTimer(_HookTimer):
    """Timer that never triggers."""
    def should_trigger_for_step(self, step):
        _ = step
        return False
    def update_last_triggered_step(self, step):
        _ = step
        return (None, None)
    def last_triggered_step(self):
        return None
''' Adapted LogTrainerHook to work with Comet
'''
class CometSessionHook(tf.train.SessionRunHook):
    def __init__(self,
                 tensors,
                 parameters,
                 every_n_iter=1,
                 every_n_secs=None,
                 at_end=False,
                 formatter=None):
        self.parameters = parameters
        only_log_at_end = (
                at_end and (every_n_iter is None) and (every_n_secs is None))
        if (not only_log_at_end and
                (every_n_iter is None) == (every_n_secs is None)):
            raise ValueError(
                "either at_end and/or exactly one of every_n_iter and every_n_secs "
                "must be provided.")
        if every_n_iter is not None and every_n_iter <= 0:
            raise ValueError("invalid every_n_iter=%s." % every_n_iter)
        if not isinstance(tensors, dict):
            self._tag_order = tensors
            tensors = {item: item for item in tensors}
        else:
            self._tag_order = sorted(tensors.keys())
            self._tensors = tensors
            self._formatter = formatter
            self._timer = (
                NeverTriggerTimer() if only_log_at_end else
                SecondOrStepTimer(every_secs=every_n_secs,
                                  every_steps=every_n_iter))
            self._log_at_end = at_end
    def begin(self):
        self.experiment = Experiment(
            api_key="API_KEY",
            project_name="estimator-hook",
            workspace="gidim")
        self.experiment.log_multiple_params(self.parameters)
        self._timer.reset()
        self._iter_count = 0
        # Convert names to tensors if given
        self._current_tensors = {tag: _as_graph_element(tensor)
                                 for (tag, tensor) in self._tensors.items()}
    def after_create_session(self, session, coord):
        self.experiment.set_model_graph(session.graph)
    def before_run(self, run_context):  # pylint: disable=unused-argument
        self._should_trigger = self._timer.should_trigger_for_step(
            self._iter_count)
        if self._should_trigger:
            # set up feed dict for sess run
            return SessionRunArgs(self._current_tensors)
        else:
            return None
    def _log_tensors(self, tensor_values):
        # log comet metrics
        self.experiment.log_multiple_metrics(tensor_values)
        original = np.get_printoptions()
        np.set_printoptions(suppress=True)
        elapsed_secs, _ = self._timer.update_last_triggered_step(
            self._iter_count)
        if self._formatter:
            logging.info(self._formatter(tensor_values))
        else:
            stats = []
            comet_stats = {}
        for tag in self._tag_order:
            stats.append("%s = %s" % (tag, tensor_values[tag]))
            comet_stats[tag] = tensor_values[tag]
        if elapsed_secs is not None:
            logging.info("%s (%.3f sec)", ", ".join(stats), elapsed_secs)
        else:
            logging.info("%s", ", ".join(stats))
            np.set_printoptions(**original)
    def after_run(self, run_context, run_values):
        _ = run_context
        if self._should_trigger:
            self._log_tensors(run_values.results)
        self._iter_count += 1
    def end(self, session):
        if self._log_at_end:
            values = session.run(self._current_tensors)
            self._log_tensors(values)
def _as_graph_element(obj):
    """Retrieves Graph element."""
    graph = ops.get_default_graph()
    if not isinstance(obj, six.string_types):
        if not hasattr(obj, "graph") or obj.graph != graph:
            raise ValueError("Passed %s should have graph attribute that is equal "
                             "to current graph %s." % (obj, graph))
        return obj
    if ":" in obj:
        element = graph.as_graph_element(obj)
    else:
        element = graph.as_graph_element(obj + ":0")
        # Check that there is no :1 (e.g. it's single output).
        try:
            graph.as_graph_element(obj + ":1")
        except (KeyError, ValueError):
            pass
        else:
            raise ValueError("Name %s is ambiguous, "
                             "as this `Operation` has multiple outputs "
                             "(at least 2)." % obj)
    return element
```

fprost:
Awesome, it should be easy with those hooks indeed.

PR #153, opened 2018-09-11
Word + Label Embeddings

jjtan:
Base word label embedding implementation.

### DIFF
--- a/experiments/tf_trainer/common/tfrecord_input.py
+++ b/experiments/tf_trainer/common/tfrecord_input.py
@@ -18,16 +18,16 @@ class TFRecordInput(dataset_input.DatasetInput):
   Handles parsing of TF Examples.
   """
 
-  def __init__(
-      self,
-      train_path: str,
-      validate_path: str,
-      text_feature: str,
-      labels: Dict[str, tf.DType],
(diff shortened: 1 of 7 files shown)


experiments/tf_trainer/common/tfrecord_input.py
```diff
@@ -50,42 +51,40 @@ def _input_fn_from_file(self, filepath: str) -> types.FeatureAndLabelTensors:
...
-    padded_shapes=(
-            {self._text_feature: [None],
-            'sequence_length': []},
-            {label: [] for label in self._labels})
+    padded_shapes = ({
+        self._text_feature: [None],
+        'sequence_length': []
+    }, {label: [] for label in self._labels})
     # TODO(fprost): Set a filter to remove long sentences (for training).
+    parsed_dataset = parsed_dataset.filter(
```
jjtan: @fprost is this roughly what you were thinking? This doesn't allow different options for training vs not but I need something to reduce seq length for now.

fprost: Yes, it is. Thank you very much for your patch, you can delete my TODO!

PR #157, opened 2018-09-24
Fix "KeyError: 'predict'" in model_trainer.py

ianchou:
Existing code return: "KeyError: 'predict'"
It seems we can't assume there is always 'predict' in export_outputs


nthain:
@fprost: Hi Flavien, could you do a quick scan of this PR? I think you've got the most familiarity with this piece.

fprost:
I recognize that previous code was raising an error with keras.
Seems that the new one is raising one on TensorFlow. Message error is: AttributeError: 'ClassificationOutput' object has no attribute 'outputs'

Any way to reconcile both? Otherwise, we might have to pass as an argument which function to use depending on the model.

Also, I submitted a bug internally but have not heard back yet.

PR #160, opened 2018-09-26
Adding loading option

fprost:
Restructuring the code to add one feature "Loading". 

Feature description:
When we call the method dataset.add_model_prediction_to_data(model) we now have an extra argument 'recompute_predictions'.
If set to False, we reload past prediction instead of rerunning the entire pipeline.

Design changes:
- Clarification of the file names. A dataset has now a 'dataset_dir' where all the files will be saved (input_tf_records and output_predictions).
- Adding good tests for the entire pipeline in dataset_test.py
The rest are minor changes to keep a clean flow.

Thank you for the review.

### DIFF
--- a/model_evaluation/utils_export/dataset_test.py
+++ b/model_evaluation/utils_export/dataset_test.py
@@ -20,6 +20,7 @@ from __future__ import print_function
 
 import getpass
 import os
+import time
 import unittest
 
 from dataset import Dataset
@@ -140,6 +141,8 @@ class TestModelCompatibleWithInputFn(unittest.TestCase):
 class TestEndPipeline(unittest.TestCase):
   """Verifies end-to-end use of dataset."""
 
+  test_version = str(int(time.time()))
+
   def setUp(self):
     def input_fn_test(max_n_examples):
       return pd.DataFrame({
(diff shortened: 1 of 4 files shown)


fprost:
Also, as soon as this is merged, I will create the README to describe all of this

model_evaluation/utils_export/dataset_test.py
```diff
@@ -91,55 +91,92 @@ def testBadTypeFeatureKeys(self):
...
+  def testComputePredictions(self):
+    try:
+      self.dataset.add_model_prediction_to_data(self.model) 
+    except ValueError :
+      self.fail('Dataset raised an exception unexpectedly!')
+
+  def testLoadPredictions(self):
+    try:
+      self.dataset.add_model_prediction_to_data(
+          self.model, recompute_predictions=False)
```
nthain: Is there an assert we can use to check that the predictions were loaded and not recomputed?

fprost: Would that make sense to delete test directory before? It would ensure that predictions are not loaded.
However, There is the risk that this directory contains valuable information (limited risk if the test directory has a "unusual enough" name)

fprost: Modified by adding a version.

PR #166, opened 2018-10-03
Use new dataset

fprost:
Starting to use civil comments dataset.

Writing input_fn in the evaluation notebook: It requires adding the tf-records spec and using decode_tf_records_to_pandas utilities.

One work item was to filter out irrelevant examples from input_fn. This is why I added a filter_fn.


model_evaluation/utils_export/utils_tfrecords.py
```diff
@@ -114,14 +114,18 @@ def encode_pandas_to_tfrecords(df, feature_keys_spec, tf_records_path, example_k
...
-    tf_records_path:
-    max_n_examples:
-    random_filter_keep_rate:
+    decoding_features_spec: A dict mapping feature keys to FixedLenFeature values.
+      Spec of the tf-records.
+    tf_records_path: path to the file
+    max_n_examples: Maximum number of examples to extract.
+    random_filter_keep_rate: Probability for each line to be kept in training data.
+      For each line, we generate a random number x and keep it if x < random_filter_keep_rate.
+    filter_fn: Function applied to an example. If return False, the example will be discarded.
```
nthain: -> filter_fn (optional):

nthain: 'If return False' -> 'If it return False'

model_evaluation/jigsaw_evaluation_pipeline.ipynb
```diff
@@ -293,6 +302,109 @@
     "    return res"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Civil bias dataset"
```
nthain: Civil comments bias dataset

model_evaluation/jigsaw_evaluation_pipeline.ipynb
```diff
@@ -293,6 +302,109 @@
...
+    "identity_terms_civil = ['male', 'female', 'transgender', 'other_gender', 'heterosexual', \n",
+    "                        'homosexual_gay_or_lesbian', 'bisexual', 'other_sexual_orientation',\n",
+    "                        'christian', 'jewish', 'muslim', 'hindu', 'buddhist', 'atheist',\n",
+    "                        'other_religion', 'black', 'white', 'asian', 'latino', 'other_race_or_ethnicity',\n",
+    "                        'physical_disability', 'intellectual_or_learning_disability',\n",
+    "                        'psychiatric_or_mental_illness', 'other_disability'\n",
+    "                       ]\n",
+    "\n",
+    "def filter_fn_civil(example, background_filter_keep_rate=0.1):\n",
+    "    if example ['male'] < 0.:\n",
```
nthain: How can example['male'] be < 0? Should this be <= 0 ?

fprost: If you look at the way I parsed the tf-records:
'male': tf.FixedLenFeature([], dtype=tf.float32, default_value=-1.)

Because male is an optional field, I need to provide a default_value. When male is not in the record, it will take the default value. I have not heard of any way to say "Parse only existing fields".

So taking ('male' <0) means we remove the lines where there was no identity information. From what I have seen, information is provided for either all entities or none. So (male>=0) means that we have information for all entities.

Does that make sense?

nthain: Yes, thanks for clarifying!

PR #167, opened 2018-10-04
Add README to model_evaluation

ianchou:


model_evaluation/jigsaw_evaluation_pipeline.ipynb
```diff
@@ -242,7 +242,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "!pip install -U -q git+https://github.com/conversationai/unintended-ml-bias-analysis@1de676a31de9e43892964f71d1e38e90fc8b331e"
```
fprost: I think this will break in the second part of the notebook. I am working on using new metrics, so let's hold on that change for now

ianchou: FWIW, the whole notebook works for me with this change.
Anyway, looking forward to your changes on using new metrics :)

PR #169, opened 2018-10-04
using new eval functions

fprost:
This commit includes the new eval function (use latest model_bias_analysis file).

Now the notebook 'jigsaw_evaluation_pipeline.ipynb' allows you to:
- select a performance dataset: can be either toxicity 2017 or civil_comments.
- select a bias dataset: can be either synthetic dataset or civil_comments (with a filter).
- Run predictions (no change).
- Run eval.


model_evaluation/jigsaw_evaluation_pipeline.ipynb
```diff
@@ -210,31 +206,111 @@
...
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "# Define features\n",
+    "TOXICITY_PERFORMANCE_DATASET = 'gs://kaggle-model-experiments/resources/toxicity_q42017_test.tfrecord'\n",
+    "LABEL_NAME_TOXICITY_TEST_FILE = 'frac_neg' #Name of the label in the performance dataset\n",
```
nthain: This could be simplified to just "TOXICITY_DATA_LABEL"

model_evaluation/jigsaw_evaluation_pipeline.ipynb
```diff
@@ -210,31 +206,111 @@
...
+   "source": [
+    "def input_fn_performance_civil(max_n_examples=None, random_filter_keep_rate=1.0):\n",
+    "    civil_df_raw = utils_tfrecords.decode_tf_records_to_pandas(\n",
+    "        civil_comments_spec,\n",
+    "        CIVIL_COMMENTS_PATH,\n",
+    "        max_n_examples=max_n_examples,\n",
+    "        random_filter_keep_rate=random_filter_keep_rate,\n",
+    "    )\n",
+    "    civil_df_raw[TEXT_FEATURE_NAME] = list(map(tokenizer, civil_df_raw[TEXT_FEATURE_NAME]))\n",
+    "    civil_df_raw['toxicity'] = list(map(lambda x :bool(round(x)), list(civil_df_raw['toxicity'])))\n",
```
nthain: I think the spacing after lambda x is wrong? Shouldn't it be 'lambda x: bool....'?

model_evaluation/jigsaw_evaluation_pipeline.ipynb
```diff
@@ -730,12 +753,12 @@
    "source": [
     "for model_family in MODEL_FAMILIES:\n",
     "  auc_list = []\n",
-    "  for model in model_family:\n",
+    "  for _model in model_family:\n",
     "    fpr, tpr, thresholds = metrics.roc_curve(\n",
     "        test_performance_df['label'],\n",
-    "        test_performance_df[model])\n",
+    "        test_performance_df[_model])\n",
```
nthain: Why is this being called _model?

model_evaluation/jigsaw_evaluation_pipeline.ipynb
```diff
@@ -826,106 +812,81 @@
...
+    "    \n",
+    "    ### Make it prettier\n",
+    "    _model = model_family[0]\n",
+    "    _nickname = model_nick_names[_model]\n",
+    "    bias_metrics_df = bias_metrics_df.rename(columns={\n",
+    "        col: col.replace(_model, _nickname) for col in bias_metrics_df.columns\n",
+    "    })\n",
+    "    for col in bias_metrics_df.columns:\n",
+    "        try:\n",
+    "            bias_metrics_df[col] = list(map(lambda x: [round(y,2) for y in x] , bias_metrics_df[col]))\n",
```
nthain: Could something like this be used instead? 
https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.round.html

model_evaluation/jigsaw_evaluation_pipeline.ipynb
```diff
@@ -210,31 +206,111 @@
...
+    "### Civil comments performance dataset"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "CIVIL_COMMENTS_PATH = 'gs://kaggle-model-experiments/resources/civil_comments_data/train.tfrecord'\n",
```
ianchou: just saying, does it make sense to train on toxicity_q42017 but test on civil_comments_data?
maybe it make more sense to do training and testing on these two dataset separately?

fprost: Good point. This evaluation notebook has now the input_fn for both of these datasets, so it is up to the user to choose the right one.
Your point is relevant and we actually will train on civil_comments from now on (at least for my experiments).

model_evaluation/jigsaw_evaluation_pipeline.ipynb
fprost: Thanks for spotting it

model_evaluation/jigsaw_evaluation_pipeline.ipynb
fprost: I tried it but it does not work. The dataframe returned contains list of values (e.g. df[col][i] = [0.897123]).
df.round(2) works only if df[col][i] = 0.89.

model_evaluation/jigsaw_evaluation_pipeline.ipynb
fprost: I changed the variable name of the for loop from model to _model, so that it does not overwrite the global variable 'model'

model_evaluation/jigsaw_evaluation_pipeline.ipynb
nthain: Thanks for catching this @ianchou!

PR #172, opened 2018-11-07
Update bias eval with newly refactor metrics code

lucyvasserman:


model_evaluation/jigsaw_evaluation_pipeline.ipynb
```diff
@@ -341,8 +349,39 @@
...
+       "True"
+      ]
+     },
+     "execution_count": 13,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "nltk.download('punkt')"
```
fprost: Thanks for adding it!

I would move this on top of the file since it is a setting for the entire notebook.

model_evaluation/jigsaw_evaluation_pipeline.ipynb
```diff
@@ -503,11 +558,20 @@
...
+     "output_type": "stream",
+     "text": [
+      "INFO:tensorflow:Model is compatible with the `Dataset` instance.\n",
+      "WARNING:tensorflow:Using past predictions. the data must match exactly (same number of lines and same order).\n"
+     ]
+    }
+   ],
    "source": [
-    "dataset_performance.add_model_prediction_to_data(model)"
+    "dataset_performance.add_model_prediction_to_data(model, recompute_predictions=False)"
```
fprost: I am wondering if we want to leave this parameter as False or True. False avoids extra computation (in particular for someone that does not know the parameter) but it would also lead to an error when using a new model.

Please do what you think is best!

PR #176, opened 2018-11-09
Basic travis config for running build and run tests with bazel.

sorensenjs:
Not tested, because needs to be submitted for test IIUC.


sorensenjs:
Yay! working, test that is failing is really failing.

PR #179, opened 2018-11-11
Adds an initial CNN specification parser

iislucas:
This is the first step of updating the CNN to use a specification parser so we can more easily experiment with various kinds of CNNs.

### DIFF
--- a/experiments/tf_trainer/common/cnn_spec_parser.py
+++ b/experiments/tf_trainer/common/cnn_spec_parser.py
@@ -1,0 +1,71 @@
+"""CNN Specification Parser
+
+Parsers the specification of convolutional layers (the number and sizes of
+filters)
+
+```
+  layers = layer : layers
+  layer = filters
+  filters = filter, filters
+  filter = numfilters * (size / stride -> output_embedding_size)
+  size, stride, output_embedding_size = \d+
+```
+"""
+
+import re
+from typing import List
+
+filter_regexp = re.compile(
+  r'(?P<num_filters>\d+)\s*\*\s*'
+  r'\(\s*(?P<size>\d+)\s*/\s*(?P<stride>\d+)\s*'
+  r'\-\>\s*(?P<output_embedding_size>\d+)\s*\)')
+
+layers_split_regexp = re.compile(r'\s*:\s*')
+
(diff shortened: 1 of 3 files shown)


experiments/tf_trainer/common/cnn_spec_parser.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class FilterParseError(Exception):
+  pass
+
+
+class Filter(object):
+  """A single CNN filter.
+
+  filter = '(size / stride -> num_filters)'
```
nthain: Why did we choose this notation 'size / stride -> num_filters'? Is there a standard or some reasoning behind it? Perhaps we can add some more information to the PR description to clarify the syntax.

iislucas: I don't know of a standard. I wanted something that I could interpret somehow... here was my thinking... 

The num filters is the final output dimension size, so I put it on the right of the arrow.
(input-length * size / stride) give you the other dimension for the matrix. 
I wanted to have some kind of expression of the input going to the output. 

I think this will have to change a little as I add max-pool vs avg-pool vs other aggregation functions. Maybe "a{ size / stride -> num filters }", or something. 

I'll add a little explanation for this. Its quite easy to change if we come up with a better notation.

sorensenjs:
Fwiw I oppose the attempt to write a custom domain specific language, like
the one we already used in atc for this purpose.

iislucas:
Yeah, writing domain specific languages is rarely good. Curious for this case how do suggest to do a series of tests with a wide variety of CNN architectural differences? I don't really want to write 50 different model files... This provides a pretty simple way to push some of the architectural experiments into h-params.

sorensenjs:
Maybe we can use a CNN to map a natural language description of the desired architecture to a python function?

iislucas:
hehe :)

sorensenjs:
This commit added f-strings which are python 3.6 features. Google
workstations are at 3.5.3. Why do you hate us?

iislucas:
oops; I did it because they are nice and didn't notice and didn't realize workstations were behind. I probably accidentally updated my workstation a while ago... I'll reformat them to be uglier...

sorensenjs:
Already done! No worries

PR #181, opened 2018-11-11
add initial converter for jsonlines to tfrecord

iislucas:
Jsonlines are generally a much better format than CSV, they also allow, in future streaming, and we have lots of datasets in jsonlines that are not in CSV.


experiments/tools/convert_jsonl_to_tfrecord.py
```diff
@@ -0,0 +1,141 @@
...
+
+class MisingAllTextFieldsError(Exception):
+  pass
+
+
+class FieldsCounter():
+  def __init__(self):
+    self.counters = {}
+
+  def inc_field(self, field_name:str):
```
nthain: This seems to be the only place where a type hint is provided. Should we make the file consistent in either direction?

experiments/tools/convert_jsonl_to_tfrecord.py
```diff
@@ -0,0 +1,141 @@
...
+      self.counters[field_name] = 0
+    self.counters[field_name] += 1
+
+
+def make_selected_output_row(row, line, counters):
+  """Create an output row with just the fields matching --text_fields_re and
+  --label_fields_re. If there is no matching field in the row for
+  --text_fields_re then raise MisingAllTextFieldsError.
+  """
+  text_field_matcher = re.compile(FLAGS.text_fields_re)
```
nthain: Will get an error if the text_fields_re is left at the default value 'None'. However, in the description of the flag you make it seem like this flag is optional.

experiments/tools/convert_jsonl_to_tfrecord.py
```diff
@@ -0,0 +1,141 @@
...
+  label_field_matcher = re.compile(FLAGS.label_fields_re)
```
nthain: Will get an error if the label_fields_re is left at the default value 'None'. However, in the description of the flag you make it seem like this flag is optional.

experiments/tools/convert_jsonl_to_tfrecord.py
```diff
@@ -0,0 +1,141 @@
...
+  'Matcher for names of the text fields. Used to compute stats if provided.')
+flags.register_validator(
+  'text_fields_re', lambda value: isinstance(value, str) and re.compile(value),
+  message='--text_field_re must be a regexp string.')
+
+flags.DEFINE_string(
+  'label_fields_re', None,
+  'Matcher for names of the label fields. Used to compute stats if provided.')
+flags.register_validator(
+  'label_fields_re', lambda value: isinstance(value, str) and re.compile(value),
```
nthain: Will fail the validator if left at the default value 'None' but the flag description makes it seem like the flag is optional.

experiments/tools/convert_jsonl_to_tfrecord.py
```diff
@@ -0,0 +1,141 @@
...
+  message='--label_fields_re must be a regexp string.')
+
+flags.DEFINE_string(
+  'input_jsonlines_path', None,
+  'Path to the JSON-lines input file.')
+flags.register_validator(
+  'input_jsonlines_path', lambda value: isinstance(value, str),
```
nthain: Will fail the validator if left at the default value 'None' but the flag description makes it seem like the flag is optional.

experiments/tools/convert_jsonl_to_tfrecord.py
```diff
@@ -0,0 +1,141 @@
...
+FLAGS = flags.FLAGS
+
+flags.DEFINE_string(
+  'text_fields_re', None,
+  'Matcher for names of the text fields. Used to compute stats if provided.')
+flags.register_validator(
+  'text_fields_re', lambda value: isinstance(value, str) and re.compile(value),
+  message='--text_field_re must be a regexp string.')
+
+flags.DEFINE_string(
```
nthain: What is the point of distinguishing between text and label fields? They don't seem to be treated differently in the code.

PR #182, opened 2018-11-11
Pull out the embedding index computation and test it

iislucas:
This PR pulls out the computation about the embedding index from the text preprocessor, provide test data, and tests it.

It also removes some unused slightly confusing code-paths from the embedding index computation: 
* It no longer has a hack to remove a header column from the embedding file. 
* Unused argument to limit the max words of the index is no longer provided. 
* Removed command line argument for one run file to use binary mode when reading the embedding file. 

There may have been reasons I didn't know for these things; do let me know if for some reason we need them, and I'll add them back in, test them, and document the reasoning. :)


experiments/tf_trainer/common/text_preprocessor.py
```diff
@@ -19,26 +34,27 @@
...
 
-  Note: Due to the lack of text preprocessing functions in tensorflow, we expect 
-    that the text is already preprocessed (list of words) in inference.
-    In training, due to the availability of tf.py_func, we handle the preprocessing.
+  Note: Due to the lack of text preprocessing functions in tensorflow, we expect
+  that the text is already preprocessed (list of words) in inference. In
+  training, due to the availability of tf.py_func, we can handle the
+  preprocessing.
+
+  The reasoning for this comes from this observation: The embedding step can be
```
fprost: I am not sure to understand
Is the "word to index" the vocabulary (word--> integer)? In this case, what is the limitation around the max_words_length?

experiments/tf_trainer/keras_gru_attention/run.py
```diff
@@ -23,8 +23,6 @@
 tf.app.flags.DEFINE_string("embeddings_path",
                            "local_data/glove.6B/glove.6B.100d.txt",
                            "Path to the embeddings file.")
-tf.app.flags.DEFINE_boolean("is_binary_embedding", False,
```
fprost: Should we remove this from all run.py (e.g. tf_gru_attention/)?

experiments/tf_trainer/keras_gru_attention/run.py
```diff
@@ -50,11 +48,10 @@ def main(argv):
   del argv  # unused
 
   embeddings_path = FLAGS.embeddings_path
-  is_binary_embedding = FLAGS.is_binary_embedding
   text_feature_name = FLAGS.text_feature_name
   key_name = FLAGS.key_name
 
-  preprocessor = text_preprocessor.TextPreprocessor(embeddings_path, is_binary_embedding)
```
fprost: Same here.

experiments/tf_trainer/common/text_preprocessor.py
fprost: I am not sure to understand.  Is "Word to index" the vocabulary as word --> integer? why is this the limiting factor and how can the max word length solve the issue?

Also, is it related to the comment above ("due to the lack...")? The problem I had was related to the string preprocessing and in particular the tokenization.

experiments/tf_trainer/keras_gru_attention/run.py
iislucas: Good point. done.

experiments/tf_trainer/keras_gru_attention/run.py
iislucas: done.

experiments/tf_trainer/common/text_preprocessor.py
iislucas: good question, I think this is an old bit of text, and is more confusing and not very helpful compared to the comment above. So I'm removing for now. :)

iislucas:
Thanks! PTAL.

PR #184, opened 2018-11-11
Organize sha256 file into subdir and update travis file

iislucas:
Moves the sha file for the blaze install into a named subdirectory so its easier to see what it's for, and updates the .travis.yaml file to use it (and updates Travis to test on python 3.6, and use streamed output for blaze so we can see errors on Travis site)


sorensenjs:
Please review the travis.yml file. This is needed to authenticate a file
pulled from a repository.

iislucas:
Ah, so it is! Thanks! I think maybe I'll move it into a directory for integration testing config, as I think that will help explain why there is an otherwise random looking hash file in the root level. Sound ok?

iislucas:
Moved in this PR; PTAL. Thanks!

PR #185, opened 2018-11-12
Removing comet in setup and updating keras_cnn

fprost:
This PR removes the remaining comet dependencies. Please let me know if I missed anything

Note: keras_cnn part was a little behind, I updated it and it is training now.


fprost:
You can see this [merge](https://github.com/conversationai/conversationai-models/pull/150) where we removed the comet. 

The explanation were a lil sparse, so I added an extra comment in the bottom.

iislucas:
Thanks! that's super helpful! 

OK, we remove comet ML and think about how to add it again later.

experiments/tf_trainer/keras_cnn/run.ml_engine.sh
```diff
@@ -4,14 +4,11 @@ GCS_RESOURCES="gs://kaggle-model-experiments/resources"
...
-LOCAL_COMET_API_KEY_FILE="comet_api_key.txt"
-REMOTE_COMET_API_KEY_FILE=${GCS_RESOURCES}/${USER}_comet_api_key.txt
 
-gsutil cp ${LOCAL_COMET_API_KEY_FILE} ${REMOTE_COMET_API_KEY_FILE} &&
-gsutil acl set private ${REMOTE_COMET_API_KEY_FILE} &&
 gcloud ml-engine jobs submit training tf_trainer_${MODEL_NAME}_${USER}_${DATETIME} \
     --job-dir=${JOB_DIR} \
-    --runtime-version=1.8 \
+    --runtime-version=1.10 \
+    --python-version "3.5" \
```
iislucas: Probably better to be using python 3.6 no ?

PR #186, opened 2018-11-12
Fix tf hub trainer - need None correct size determination.

sorensenjs:


fprost:
I am not sure about the shapes here.

In your current _read_tf_example [https://github.com/conversationai/conversationai-models/blob/master/experiments/tf_trainer/common/tfrecord_simple.py#L81](url), you are converting the label to a singleton list (instead of a scalar) and  this is why you need to use padded_shape = None (instead of [] for scalar).

I am not sure if this is compatible. I believe that your labels would be then 2d for your model function whereas they are typically 1-D. Have you been able to see if the model is training correctly now.

I can run some tests later today.

sorensenjs:
Ah, okay, this dataset code is  still largely a mystery to me, but I think I've fixed this and now it's closer to the code in tfrecord_input. (My goal is to make tfrecord_simple be the base class.)

PR #187, opened 2018-11-13
Cleaning tf record simple

fprost:
Fixing the size problem:
- In serving function: `input_example_tensor` expects a batch of examples (1D), so fixing it to None.
- In the input function: in our case, the batching is very simple (just concatenating scalar of strings (features) and scalar of integers (labels). 
Padding is not required (padding is when we split the strings into different words before hand. In this case, each sentence has different length and we need to pad them). 

Code runs now and exports. I have not ran a long training job though.


fprost:
Jeffrey, I have changed the tests to expect scalar instead a list of a single value. Let me know if it makes sense!

PR #188, opened 2018-11-15
Refactor comment_key pass thru feature

sorensenjs:
Remove the signature for the input comment_key int64 feature, making it optional.


sorensenjs:
I refactored this again, preserving the feature but pushing the flag closer to it's only use.
I've also set it with a default.
Sorry that tf_hub classifier duplicates the serving function, but it has no vocabulary.

PR #190, opened 2018-11-18
A major refactoring of the input system.

sorensenjs:
Yes, this still uses lots of flags, but has decoupled mostly the
input system from the model. (There are still a few hard coded
'comment_text' feature bindings in the keras directories.)

The decision to hard code the 'text' feature in the simple case
as is used in tf_hub is a reflection of the way our analyzers are
currently coded.

There are several fixes to things that are currently broken,
and all of these models have been tested locally.

### DIFF
--- a/experiments/tf_trainer/common/tfrecord_input.py
+++ b/experiments/tf_trainer/common/tfrecord_input.py
@@ -8,43 +8,34 @@ import multiprocessing
 
 import tensorflow as tf
 from tf_trainer.common import dataset_input
+from tf_trainer.common import tfrecord_simple
 from tf_trainer.common import types
+
 from typing import Callable, Dict, List
 
 
-class TFRecordInput(dataset_input.DatasetInput):
+class TFRecordInput(tfrecord_simple.TFSimpleRecordInput):
   """TFRecord based DatasetInput.
 
   Handles parsing of TF Examples.
   """
 
   def __init__(self,
-               train_path: str,
-               validate_path: str,
-               text_feature: str,
-               labels: Dict[str, tf.DType],
                train_preprocess_fn: Callable[[str], List[str]],
                batch_size: int = 64,
-               round_labels: bool = True,
                num_prefetch: int = 5,
                max_seq_len: int = 30000) -> None:
-    self._train_path = train_path
-    self._validate_path = validate_path
-    self._text_feature = text_feature
-    self._labels = labels
-    self._batch_size = batch_size
+    super().__init__(batch_size, num_prefetch)
     self._train_preprocess_fn = train_preprocess_fn
-    self._round_labels = round_labels
-    self._num_prefetch = num_prefetch
     self._max_seq_len = max_seq_len
 
-  def train_input_fn(self) -> types.FeatureAndLabelTensors:
-    """input_fn for TF Estimators for training set."""
-    return self._input_fn_from_file(self._train_path)
+  def tokens_feature(self) -> str:
+    """Returns the name of the feature containing integer tokens.
 
-  def validate_input_fn(self) -> types.FeatureAndLabelTensors:
-    """input_fn for TF Estimators for validation set."""
-    return self._input_fn_from_file(self._validate_path)
+    Note that currently this code overwrites the input text feature
+    with the tokens using the same name.
+    """
+    return self.text_feature()
 
   def _input_fn_from_file(self, filepath: str) -> types.FeatureAndLabelTensors:
 
@@ -54,13 +45,13 @@ class TFRecordInput(dataset_input.DatasetInput):
 
     parsed_dataset = dataset.map(
         self._read_tf_example, num_parallel_calls=multiprocessing.cpu_count())
+    parsed_dataset = parsed_dataset.filter(
+        lambda x, _: tf.less(x['sequence_length'], self._max_seq_len))
 
     padded_shapes = ({
-        self._text_feature: [None],
+        self.tokens_feature(): [None],
         'sequence_length': []
     }, {label: [] for label in self._labels})
-    parsed_dataset = parsed_dataset.filter(
-        lambda x, _: x['sequence_length'] < self._max_seq_len)
     parsed_dataset = parsed_dataset.apply(
         tf.contrib.data.bucket_by_sequence_length(
             element_length_func=lambda x, _: x['sequence_length'],
@@ -79,19 +70,19 @@ class TFRecordInput(dataset_input.DatasetInput):
     The input TF Example has a text feature as a singleton list with the full
     comment as the single element.
     """
-    DEFAULT_VALUES = {tf.string: '', tf.float32: -1.0, tf.int32: -1}
-
     keys_to_features = {}
-    keys_to_features[self._text_feature] = tf.FixedLenFeature([], tf.string)
-    for label, dtype in self._labels.items():
-      keys_to_features[label] = tf.FixedLenFeature([], dtype,
-                                                   DEFAULT_VALUES[dtype])
+    keys_to_features[self.text_feature()] = tf.FixedLenFeature([], tf.string)
+    for label in self._labels:
+      keys_to_features[label] = tf.FixedLenFeature([], tf.float32, -1.0)
     parsed = tf.parse_single_example(
         record, keys_to_features)  # type: Dict[str, types.Tensor]
 
-    text = parsed[self._text_feature]
-    features = {self._text_feature: self._train_preprocess_fn(text)}
-    features['sequence_length'] = tf.shape(features[self._text_feature])[0]
+    text = parsed[self.text_feature()]
+    tokens = self._train_preprocess_fn(text)
+    features = {
+        self.tokens_feature(): tokens,
+        'sequence_length': tf.shape(tokens)[0],
+    }
     if self._round_labels:
       labels = {label: tf.round(parsed[label]) for label in self._labels}
     else:
--- a/experiments/tf_trainer/common/tfrecord_simple.py
+++ b/experiments/tf_trainer/common/tfrecord_simple.py
@@ -10,6 +10,22 @@ import tensorflow as tf
 from tf_trainer.common import dataset_input
 from tf_trainer.common import types
 from typing import Dict
+from typing import List
+
+
+tf.app.flags.DEFINE_string('train_path', None,
+                           'Path to the training data TFRecord file.')
+tf.app.flags.DEFINE_string('validate_path', None,
+                           'Path to the validation data TFRecord file.')
+tf.app.flags.DEFINE_string('labels', 'frac_neg',
+                           'Comma separated list of (float) label features.')
+tf.app.flags.DEFINE_string('text_feature', 'comment_text',
+                           'Name of feature containing text input.')
+tf.app.flags.DEFINE_boolean('round_labels', True,
+                            'Round label features to 0 or 1 if true.')
+
+
+FLAGS = tf.app.flags.FLAGS
 
 
 class TFSimpleRecordInput(dataset_input.DatasetInput):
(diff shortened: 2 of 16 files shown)


experiments/tf_trainer/common/tfrecord_simple.py
```diff
@@ -60,17 +79,15 @@ def _read_tf_example(
...
-    for label, dtype in self._labels.items():
-      keys_to_features[label] = tf.FixedLenFeature([], dtype,
-                                                   DEFAULT_VALUES[dtype])
+    for label in self._labels:
+      keys_to_features[label] = tf.FixedLenFeature([], tf.float32, -1.0)
     parsed = tf.parse_single_example(
         record, keys_to_features)  # type: Dict[str, types.Tensor]
 
-    features = {self._text_feature: parsed[self._text_feature]}
+    features = {'text': parsed[self._text_feature]}
```
iislucas: It is probably better to make it a class static var (so we can change it in one place should we need to), avoiding lots of string literal dependencies, and perhaps it should be noted in the class doc-string that the output text feature will always be called 'text' no matter the value of `self._text_feature` ?

experiments/tf_trainer/common/tfrecord_simple_test.py
```diff
@@ -46,44 +49,35 @@ def setUp(self):
...
-        labels={"label": tf.float32},
-        round_labels=False)
+    FLAGS.round_labels = False
+    FLAGS.labels = 'label'
+    dataset_input = tfrecord_simple.TFSimpleRecordInput()
 
     with self.test_session():
       features, labels = dataset_input._read_tf_example(self.ex_tensor)
-      self.assertEqual(features["comment"].eval(), b"Hi there Bob")
+      self.assertEqual(features["text"].eval(), b"Hi there Bob")
```
iislucas: why not use single quotes like everywhere else?

experiments/tf_trainer/common/tfrecord_simple.py
```diff
@@ -10,6 +10,22 @@
 from tf_trainer.common import dataset_input
 from tf_trainer.common import types
 from typing import Dict
+from typing import List
+
+
+tf.app.flags.DEFINE_string('train_path', None,
```
iislucas: I actually think this part of the PR change - of making these global re-used flags - is net worse for the codebase. I know it puts some common code in one place, but it pushes the flag var definitions throughout many files, so to know the flags you have to do more search, and it's not obvious from the run files how far you have to dig, and what you have to provide. If we want to abstract these, I think it's best to do it into a file called `common_arguments.py` or something like that, so that it is directly included in the run files. What do you think?

experiments/tf_trainer/tf_word_label_embedding/run.py
```diff
@@ -34,35 +34,25 @@
...
-      text_feature=text_feature_name,
-      labels=LABELS,
       train_preprocess_fn=train_preprocess_fn,
       batch_size=FLAGS.batch_size,
       max_seq_len=5000)
 
   model_tf = tf_word_label_embedding.TFWordLabelEmbeddingModel(
-      text_feature_name, "frac_neg")
-  model = preprocessor.add_embedding_to_model(model_tf, text_feature_name)
+      dataset.tokens_feature(), "frac_neg")
```
iislucas: single quotes nit. :)

experiments/tf_trainer/common/tfrecord_input.py
```diff
@@ -8,43 +8,34 @@
...
-  def train_input_fn(self) -> types.FeatureAndLabelTensors:
-    """input_fn for TF Estimators for training set."""
-    return self._input_fn_from_file(self._train_path)
+  def tokens_feature(self) -> str:
+    """Returns the name of the feature containing integer tokens.
 
-  def validate_input_fn(self) -> types.FeatureAndLabelTensors:
-    """input_fn for TF Estimators for validation set."""
-    return self._input_fn_from_file(self._validate_path)
+    Note that currently this code overwrites the input text feature
```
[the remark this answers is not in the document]
sorensenjs: Yes, I want to fix this and tried to create a new 'tokens' feature that we use instead of editing the content (and type!) of the feature. But: it's hard coded in the keras code to use the 'comment_text' feature
https://github.com/conversationai/conversationai-models/blob/master/experiments/tf_trainer/keras_cnn/model.py#L67
one step at a time?

experiments/tf_trainer/common/tfrecord_input.py
```diff
@@ -8,43 +8,34 @@
...
 
   Handles parsing of TF Examples.
   """
 
   def __init__(self,
-               train_path: str,
-               validate_path: str,
-               text_feature: str,
-               labels: Dict[str, tf.DType],
                train_preprocess_fn: Callable[[str], List[str]],
```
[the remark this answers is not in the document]
sorensenjs: This change, while not a bad idea, is immaterial to this change. I will argue that a python text preprocessing function does not belong in either the input class or the model class. Getting this abstraction right is why we have 4 different projects and team working on this.

experiments/tf_trainer/common/tfrecord_simple.py
sorensenjs: Sending a separate email to explain the use of 'text'

experiments/tf_trainer/common/tfrecord_simple.py
sorensenjs: Why I disagree: the flags are now only scoped to this single file, they are no longer globally re-used, or at lest that is my intention. That is why I introduced accessors to the input classes to catch attempts to access flag contents.

That makes it possible to introduce a new abstraction that maps say 'civil_comments' -> input files, validation file, label keys etc.

But: that can't be done at the same step because we have about 12 shell scripts that all specify flags.

This is a behavior preserving change. You suggestion would broaden the scope and require changes to multiple files for which there is no test coverage. More importantly, the idea of common_arguments belies the more fundamental problem with our abstractions that two file names and a list of strings are inadequate to describe more than a subset of potential inputs.

experiments/tf_trainer/tf_word_label_embedding/run.py
sorensenjs: quoting only has to be consistent within file - but ugh, doing it.

experiments/tf_trainer/common/tfrecord_simple_test.py
sorensenjs: This is not consistent in our code, but doing it.

iislucas: Urg... you're right. OK, we'll work our way to cleaning this up.

experiments/tf_trainer/tf_word_label_embedding/run.py
iislucas: sorry. and thank you :)

experiments/tf_trainer/common/tfrecord_simple.py
iislucas: Thanks, that helps. My summary, for github, is that is that consumers of the model may have different ways they provide the model with the text feature at evaluation time. And you wanted to make it fixed to 'text' for one of them. I think it might be better to do that as a data-set pre-processing step, or provide it as a command line argument (I guess I'd name the command line argument something like "renamed_text_field")? I'm a little skeptical about this being the right place for this abstraction, but it may be convenient to be able to change the output model like this, and it's not terribly painful.

experiments/tf_trainer/tf_hub_classifier/model.py
```diff
@@ -54,7 +52,7 @@ def estimator(self, model_dir):
 
   def _model_fn(self, features, labels, mode, params, config):
     embedded_text_feature_column = hub.text_embedding_column(
-      key=self._text_feature_name, module_spec=FLAGS.model_spec)
+      key='text', module_spec=FLAGS.model_spec)
```
iislucas: Does this now want to be `base_model.TEXT_FEATURE_KEY` ? 
If not, maybe explain why?

experiments/tf_trainer/tf_hub_classifier/run.py
```diff
@@ -41,7 +35,7 @@ def serving_input_fn_tfrecords():
         name="input_example_tensor"
     )
     feature_spec = {
-        text_feature_name: tf.FixedLenFeature([], dtype=tf.string),
+        'text': tf.FixedLenFeature([], dtype=tf.string),
```
iislucas: This one too.. `base_model.TEXT_FEATURE_KEY` ?

experiments/tf_trainer/common/tfrecord_simple.py
iislucas: OK, discussed and agreed that we want examples to clearly explain usage as we move flags out of the run.py, so that it's clear what the key flags are. And we will followup with some large improvements to structure of data input in future.

experiments/tf_trainer/common/tfrecord_simple.py
iislucas: This has now been abstracted to a variable in base_model.

experiments/tf_trainer/common/tfrecord_input.py
```diff
@@ -79,19 +70,19 @@ def _read_tf_example(
     The input TF Example has a text feature as a singleton list with the full
     comment as the single element.
```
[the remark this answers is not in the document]
iislucas: Not relevant as overwriting doesn't happen anymore; now using a named tokens feature.

experiments/tf_trainer/common/tfrecord_input.py
```diff
@@ -7,45 +7,35 @@
...
+from tf_trainer.common import base_model
 from tf_trainer.common import dataset_input
+from tf_trainer.common import tfrecord_simple
 from tf_trainer.common import types
+
 from typing import Callable, Dict, List
 
 
-class TFRecordInput(dataset_input.DatasetInput):
+class TFRecordInput(tfrecord_simple.TFSimpleRecordInput):
```
fprost: I am not a fan of the naming, as it does not highlight enough the difference between both classses.
What about:
TFSimpleRecordInput --> TFRecordInput?
TFRecordInput --> TFRecordInputWithTokenizer?

experiments/tf_trainer/common/tfrecord_simple.py
```diff
@@ -7,40 +7,64 @@
...
+                           'Name of feature containing text input.')
+tf.app.flags.DEFINE_boolean('round_labels', True,
+                            'Round label features to 0 or 1 if true.')
+
+
+FLAGS = tf.app.flags.FLAGS
 
 
 class TFSimpleRecordInput(dataset_input.DatasetInput):
   """Simple no-preprecoessing TFRecord based DatasetInput.
```
fprost: typo

experiments/tf_trainer/tf_hub_classifier/model.py
sorensenjs: Done.

experiments/tf_trainer/tf_hub_classifier/run.py
sorensenjs: Done.

experiments/tf_trainer/common/tfrecord_input.py
sorensenjs: Let's do this in a separate, rename only change.

experiments/tf_trainer/common/tfrecord_simple.py
sorensenjs: Fixed.

sorensenjs:
Nithum: there are no changes to the train procedure, so nothing in the README needs to change (at least as far as I can see.)

PR #199, opened 2018-11-26
Add feature weights to batching branch.

sorensenjs:
Use deep copy to prevent modifing feaures passed in by reference.


sorensenjs:
The deep copy is not strictly necessary, but while I was debugging I thought it was strange that the input was being changed - we create a new labels dictionary for the other output so I thought it would be a good idea to make sure not to modify the input. Keep in mind that the dictionary only has two values in it and when I copy to a new dictionary, I am copying the references to the underlying values.

PR #200, opened 2018-11-27
Moves the dataset repeat higher in the processing chain.

sorensenjs:
Does not repeat over the evaluation data.

### DIFF
--- a/experiments/tf_trainer/common/tfrecord_input.py
+++ b/experiments/tf_trainer/common/tfrecord_input.py
@@ -56,9 +56,12 @@ class TFRecordInput(dataset_input.DatasetInput):
     return self._text_feature
 
   def train_input_fn(self) -> types.FeatureAndLabelTensors:
-    """input_fn for TF Estimators for training set."""
+    """input_fn for TF Estimators for training set.
+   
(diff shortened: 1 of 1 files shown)


experiments/tf_trainer/common/tfrecord_input.py
```diff
@@ -56,9 +56,12 @@ def text_feature(self) -> str:
...
 
   def train_input_fn(self) -> types.FeatureAndLabelTensors:
-    """input_fn for TF Estimators for training set."""
+    """input_fn for TF Estimators for training set.
+   
+    Automatically repeats over input data forever.
+    """
     assert FLAGS.train_path
-    return self._input_fn_from_file(FLAGS.train_path)
+    return self._input_fn_from_file(FLAGS.train_path).repeat()
```
nthain: Should the .repeat() also be on the validate_input_fn()?

sorensenjs: No, I think to denoise the graphs we should always use the same portion of the validation set - or, actually, the entire validation set each step.

nthain: SGTM.

PR #201, opened 2018-11-27
Moveflags

sorensenjs:
Moves most of the training related flag definitions to where they are used.

Configurations should be injected via structures where possible. In this case, the properties
of the training set - the locations of the files, and the sizes of the batches are typically more
closely associated with training data itself rather than the model.

Currently we are accessing flags that are defined in duplicate in several places. This flag
duplication is very brittle. So it's preferable to define the flags where they are use to increase
encapsulation.

Eventually we will move to a model that describes all of the flags associated with a
particular training set as one structure.

### DIFF
--- a/experiments/tf_trainer/tf_word_label_embedding/run.py
+++ b/experiments/tf_trainer/tf_word_label_embedding/run.py
@@ -21,28 +21,17 @@ FLAGS = tf.app.flags.FLAGS
 tf.app.flags.DEFINE_string('embeddings_path',
                            'local_data/glove.6B/glove.6B.100d.txt',
                            'Path to the embeddings file.')
-tf.app.flags.DEFINE_integer('batch_size', 64,
-                            'The batch size to use during training.')
-tf.app.flags.DEFINE_integer('train_steps', 30000,
-                            'The number of steps to train for.')
-tf.app.flags.DEFINE_integer('eval_period', 500,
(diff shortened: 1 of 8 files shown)


experiments/tf_trainer/tf_word_label_embedding/run.py
```diff
@@ -21,28 +21,17 @@
...
-
-  preprocessor = text_preprocessor.TextPreprocessor(embeddings_path)
+  preprocessor = text_preprocessor.TextPreprocessor(FLAGS.embeddings_path)
 
   nltk.download('punkt')
   train_preprocess_fn = preprocessor.train_preprocess_fn(nltk.word_tokenize)
   dataset = tfrecord_input.TFRecordInputWithTokenizer(
       train_preprocess_fn=train_preprocess_fn,
-      batch_size=FLAGS.batch_size,
       max_seq_len=5000)
```
fprost: Should be set as a flag. Or at least should not be overwriting a default value from https://github.com/conversationai/conversationai-models/blob/cb769b8a786b248cde6ba3f7e309a7a14dc05a47/experiments/tf_trainer/common/tfrecord_input.py#L146

sorensenjs: Will address max_seq_len in a future change.

PR #202, opened 2018-11-30
TF CNN with dataset-specific hparams

msushkov:


experiments/tf_trainer/tf_cnn/model.py
```diff
@@ -0,0 +1,110 @@
...
+    logits = inputs
+
+    # Conv
+    X = inputs
+    for filter_size in params.filter_sizes:
+        X = layers.Conv1D(params.num_filters, filter_size, activation='relu', padding='same')(X)
+    if params.pooling_type == "average":
+      X = layers.GlobalAveragePooling1D()(X)
+    elif params.pooling_type == "max":
+      X = layers.GlobalMaxPooling1D()(X)
```
sorensenjs: else:
  raise ValueError('Unrecognized pooling type parameter')

experiments/tf_trainer/tf_cnn/model.py
```diff
@@ -0,0 +1,110 @@
...
+        model_fn=self._model_fn,
+        params=self.hparams(),
+        config=tf.estimator.RunConfig(model_dir=model_dir))
+    return estimator
+
+  def _model_fn(self, features, labels, mode, params, config):
+    inputs = features[base_model.TOKENS_FEATURE_KEY]
+    batch_size = tf.shape(inputs)[0]
+
+    logits = inputs
```
sorensenjs: if this assignment is used (if filter_sizes is None and dense_units is None) it seems degenerate. Maybe you don't need it?

msushkov:
@nthain makes it easier to run tuning jobs by calling run.hyparparameter.sh with the dataset arg. Can extend this to run.local and run.ml_engine too. After tuning is done can change run.py to use the best hparams. The idea for the per-dataset configs is that instead of needing to change dataset paths every time we want to run a script we can just pass in an argument. This may not be the cleanest way to do it, though. What are your thoughts?

experiments/tf_trainer/tf_cnn/model.py
msushkov: Thanks for catching this

nthain:
> @nthain makes it easier to run tuning jobs by calling run.hyparparameter.sh with the dataset arg. Can extend this to run.local and run.ml_engine too. After tuning is done can change run.py to use the best hparams. The idea for the per-dataset configs is that instead of needing to change dataset paths every time we want to run a script we can just pass in an argument. This may not be the cleanest way to do it, though. What are your thoughts?

Makes sense! Thanks for clarifying!

PR #205, opened 2018-12-03
Add label dtypes

nthain:
This PR allows for labels to have different dtypes via a new flag "label_dtypes". This can be used to specify the dtypes of each label as being either "float" or "int".

It also fixes:
  - typo in a local run script (sever_toxicity -> severe_toxicity)
  - inconsistent quoting in tfrecord_input.py

This branches from msushkov's commit "Adding test for int64 label" and so also includes unmerged changes from that branch.


msushkov:
LGTM

msushkov:
Thanks Nithum!

PR #211, opened 2018-12-08
Clean up run scripts; add deploy script for CNN

msushkov:


experiments/tf_trainer/tf_gru_attention/hparam_config_toxicity.yaml
```diff
@@ -1,11 +1,12 @@
 trainingInput:
   pythonVersion: '3.5'
-  scaleTier: CUSTOM
-  masterType: standard
-  workerType: standard_gpu
-  parameterServerType: large_model
-  workerCount: 1
-  parameterServerCount: 1
+  # scaleTier: CUSTOM
```
sorensenjs: maybe delete instead of commenting

Issue #217, opened 2018-12-17
JSONDecodeError in model evaluation


msushkov:
Reproduced in https://github.com/conversationai/conversationai-models/blob/msushkov_eval2/model_evaluation/model_eval.ipynb

fprost:
Could you point at the model that you are using? Is is tf_gru_attention or another one?

fprost:
I am assuming here that you are using the code for tf_gru_attention in your branch.

So looking at your notebook, it seems that:
- the API is  reading correctly your test file
- Is converting correctly to TF-Records
- Calls a batch prediction in CMLE
- BUT is unable to read back the prediction files.

Looking at your CMLE prediction job, I see that it was not able to parse the input file (otherwise you would have an example count).

This issue should come from a mismatch between the real model signature (the .pb that you deployed on CMLE) and what you described (your `Model` instance).
In particular, you are saying that:
- input spec = 'coment_text'
- example key = 'comment_key' (ok defined in model_trainer.py)
- prediction key = 'toxicity/logistic'.
Whereas from your tf_gru_attention code (hence your .pb), I see that:
- example_key is 'tokens' (in common/base_model.py called my run.py)
- prediction_key is tfrecord_input/logistic (in tfrecord_input.py). This mismatch is also due that you are training on toxicity_2017 whereas the notebook handles by default civil_comments.

So short term change to get your code to work should be to update your notebook with the right signature. Let me know if you run into issues here.

Long term change should solve:
1- Help the user to realize when the model signature does not match the on of the .pb:
- Can we automatically detect the signature of a model deployed on CMLE? In this case, we would send error messages when creating the `Model` instance.
- Otherwise, send more explicit error message when prediction file is empty.
2- Update default value of the notebook to `tokens` (made incompatible by [this commit](https://github.com/conversationai/conversationai-models/commit/0dac3ef488097163c28519c1233fd1f44a91a42b))
3- write better documentation for the `Model` attributes.

@msushkov, do you think that these changes could help avoid similar problems in the future?
 If yes, I will create a feature request and work on it when I have time.

msushkov:
Thanks @fprost! My mistake. These long-term changes would definitely make things easier, especially automatically detecting the model signature or at the very least an explicit message when the prediction file is empty (the JSONDecodeError seems confusing to debug).

fprost:
This problem was solved by the two PR #225 and #226. 

[...]

PR #226, opened 2019-01-02
Evaluation notebook readability

fprost:
Reformatted the notebook to make it easier to adapt it for other datasets or model.

Main changes are:
- Putting example of input_fn into a separate files. this makes the notebook shorter and also highlights that input_fn is easily customizable.
- Trying to highlight better the user inputs in the notebook.

Mikhail, Can you go through the notebook and let me know if you find it easier to understand. In particular, would you be more comfortable to change the dataset (input_fn) or change the model (different model signatures)?

Thank you!


fprost:
Should also contribute to solving [this issue](https://github.com/conversationai/conversationai-models/issues/217#event-2029346108).

model_evaluation/jigsaw_evaluation_pipeline.ipynb
```diff
@@ -339,7 +226,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "!pip install -U -q git+https://github.com/conversationai/unintended-ml-bias-analysis"
+    "# User inputs: Model description\n",
+    "TEXT_FEATURE_NAME = 'comment_text' #Input text\n",
+    "SENTENCE_KEY = 'comment_key' #Input key\n",
```
msushkov: Does it make sense to add a slightly more in-depth comment explaining what SENTENCE_KEY is and whether the user needs to change the default?

model_evaluation/input_fn_example.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+#TODO(fprost): Clean this file.
+
+#### #### #### #### #### ####
+#### PERFORMANCE DATASET ####
+#### #### #### #### #### ####
+
+
+def create_input_fn_toxicity_performance(tokenizer, model_input_comment_field):
```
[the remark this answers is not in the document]
fprost: Thanks for the suggestion, this file is indeed quite messy and any ideas are welcomed.

The problem that I have here is that input_fn implementation are very dependent on your test file and have a lot of variations:
- Format: civil/toxicity datasets are tfrecords (in this case, utils_tfrecords.decode_tf_records_to_pandas does most of the work) but the synthetic one is pandas.
- Fields to parse: list of identity columns.
- Transformation to make on each field: labels might have to be turned into boolean.
- Filtering, we might want to keep a subset of the examples (e.g. keep only the ones with identity or apply a random filter).
- Name of the fields can change
+ anything that another test file might have (e.g. multiple labels)

Any idea how to make it more general?

model_evaluation/jigsaw_evaluation_pipeline.ipynb
fprost: Good point, I added some comments and will resolve this issue with new PR. The first two arguments will be visible from run.py

model_evaluation/input_fn_example.py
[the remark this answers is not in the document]
msushkov: Yeah that makes sense. The only thing I can think of is adding a separate base function for each type of data (tfrecord vs pandas) and adding an arg that accepts a transformation function that will convert labels into the appropriate type. Perhaps just easier to have a separate one per dataset then, even if it means some repeated code. From a user standpoint it may not matter too much as long as it's clear in the documentation which function needs to be called.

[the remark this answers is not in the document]
fprost: Your suggestion makes sense and I agree it could be nicer.

Can I leave it as a TODO for now and we will revisit it when we have an idea if it will be common to write new/custom input_fn? Or do we want to make it a requirement for this PR?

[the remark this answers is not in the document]
fprost: Also, to your point, from docstring+ naming, is it clear which input_fn to use?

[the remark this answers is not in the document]
msushkov: Yup, I think it makes sense to revisit this later, the docs make it clear which one to use.

model_evaluation/input_fn_example.py
```diff
@@ -0,0 +1,232 @@
...
+  """Generates an input_fn to evaluate model performance on toxicity dataset."""
+
+  TOXICITY_PERFORMANCE_DATASET = 'gs://kaggle-model-experiments/resources/toxicity_q42017_test.tfrecord'
```
msushkov: Update this to use the new storage bucket

msushkov: As well as all the other paths

model_evaluation/jigsaw_evaluation_pipeline.ipynb
fprost: Updated comments.

model_evaluation/input_fn_example.py
fprost: Done. Also for the ones in notebook

PR #227, opened 2019-01-03
Example key dependency

fprost:
Moving the example_key file to the run.py. It then feeds the two files (serving_input.py and model_trainer.py) through function arguments.

This should simplify the readability of the code. In particular, we can now see the input signature of the model in run.py (they are arguments of the function create_text_serving_input_fn).


experiments/tf_trainer/keras_gru_attention/run.py
```diff
@@ -21,6 +21,8 @@
 tf.app.flags.DEFINE_string("embeddings_path",
                            "local_data/glove.6B/glove.6B.100d.txt",
                            "Path to the embeddings file.")
+tf.app.flags.DEFINE_string('key_name', 'comment_key',
```
sorensenjs: I oppose defining flags in more than one place. They are global variables (and thus pretty bad.)
My suggestion is that this properly belongs where it is, and that you expose access to it by making it a member of model_trainer (self._key_name and add an accessor.)

I think serving_input should be removed entirely and its functionality moved into model_trainer. This might also help.

sorensenjs: Sorry, flags that *share the same name* should be defined once and only once globally. I don't really care where, but probably closest to where they are used the most.

Another good principle is that flags should scope only to the file, and only to the main() function within the file (which is itself global).

fprost: In this case, in this PR, the flag redundancy is removed right? It is defined in the run.py and is used only there. Am I missing anything?

sorensenjs:
There are many run.py's. how do we know the type of each matches the
expected type where it is used?

fprost:
ok will follow your suggestion to move the serving_fn to model_trainer. It should solve the redundancy problem

msushkov:
> Added a doctring to explain it. Let me know if it makes more sense!

Yup, thanks!

PR #228, opened 2019-01-03
Refactoring run scripts and moving paths to new project

nthain:
This PR refactors the paths for the run scripts and points them to the path for the new cloud project (please refer to the e-mail I sent and respond there if there are any issues or concerns). It also reduces the default batch size of the models so that you don't by default run into OOM error when training on ML engine. 

In a future PR, I will refactor so that we will finally have one run script that can run all of the models instead of having a separate run script per model.

### DIFF
--- a/experiments/tf_trainer/common/dataset_config.sh
+++ b/experiments/tf_trainer/common/dataset_config.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+BASE_PATH="gs://conversationai-models"
+GCS_RESOURCES="${BASE_PATH}/resources"
+MODEL_PARENT_DIR="${BASE_PATH}/tf_trainer_runs"
+
+if [ "$1" == "civil_comments" ]; then
+    train_path="${GCS_RESOURCES}/civil_comments_data/train_eval_test/train-*.tfrecord"
+    valid_path="${GCS_RESOURCES}/civil_comments_data/train_eval_test/eval-*.tfrecord"
+    labels="toxicity"
+    label_dtypes="float"
+    train_steps=24000
+    eval_period=800
+    eval_steps=50
+
+elif [ "$1" == "toxicity" ]; then
+    train_path="${GCS_RESOURCES}/toxicity_data/toxicity_q42017_train.tfrecord"
+    valid_path="${GCS_RESOURCES}/toxicity_data/toxicity_q42017_validate.tfrecord"
+    labels="frac_neg"
+    label_dtypes="float"
+    train_steps=24000
+    eval_period=800
+    eval_steps=50
+
+elif [ "$1" == "many_communities" ]; then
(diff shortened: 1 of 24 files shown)


experiments/tf_trainer/tf_char_cnn/run.hyperparameter.sh
```diff
@@ -1,48 +1,10 @@
 #!/bin/bash
 
-GCS_RESOURCES="gs://kaggle-model-experiments/resources"
+source "tf_trainer/common/dataset_config.sh"
 DATETIME=`date '+%Y%m%d_%H%M%S'`
```
sorensenjs: driveby - use $() instead of backtics.

experiments/tf_trainer/common/dataset_config.sh
```diff
@@ -0,0 +1,37 @@
...
+    label_dtypes="float"
+    train_steps=24000
+    eval_period=800
+    eval_steps=50
+
+elif [ "$1" == "many_communities" ]; then
+    train_path="${GCS_RESOURCES}/transfer_learning_data/many_communities/20181105_train.tfrecord"
+    valid_path="${GCS_RESOURCES}/transfer_learning_data/many_communities/20181105_validate.tfrecord"
+    labels="removed"
+    label_dtypes="int"
```
iislucas: Removed is an integer? If so maybe explain what the interpretation of the integer is, e.g. 0 = not removed, 1 = removed.

experiments/tf_trainer/common/dataset_config.sh
```diff
@@ -0,0 +1,37 @@
...
+    train_steps=100000
+    eval_period=800
+    eval_steps=50
+
+else
+    echo "First positional arg must be one of civil_comments, toxicity, many_communities."
+    exit 1
+fi
```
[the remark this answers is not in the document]
nthain: Done.

experiments/tf_trainer/common/tfrecord_input.py
```diff
@@ -27,7 +27,7 @@
                            'Name of feature containing text input.')
 tf.app.flags.DEFINE_boolean('round_labels', True,
                             'Round label features to 0 or 1 if true.')
-tf.app.flags.DEFINE_integer('batch_size', 256,
+tf.app.flags.DEFINE_integer('batch_size', 32,
```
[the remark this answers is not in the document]
nthain: Removed from here and I'll propose it in a different PR.

experiments/tf_trainer/tf_char_cnn/run.hyperparameter.sh
nthain: Changed all occurrences.

experiments/tf_trainer/common/dataset_config.sh
nthain: Done.

PR #235, opened 2019-01-22
Updating best hyperparams for tf_cnn, tf_hub, tf_gru models.

msushkov:
Also using cached TF-Hub module.


msushkov:
@nthain did some slight cleaning up to address your comment above

PR #238, opened 2019-01-27
Update to new tf hub model, enable adapting weights.

sorensenjs:
Still training (slowly) now, but already substantially better than without adaptation of weights.

### DIFF
--- a/experiments/tf_trainer/tf_hub_classifier/myrunlocal.sh
+++ b/experiments/tf_trainer/tf_hub_classifier/myrunlocal.sh
@@ -33,7 +33,7 @@ python -m tf_trainer.tf_hub_classifier.run \
   --train_path=$train_path \
   --validate_path=$valid_path \
   --model_dir="tf_hub_classifier_local_model_dir" \
-  --model_spec="https://tfhub.dev/google/universal-sentence-encoder-lite/2" \
+  --model_spec="gs://conversationai-models/resources/tfhub/universal-sentence-ecncoder-large-3/96e8f1d3d4d90ce86b2db128249eb8143a91db73" \
   --labels=$labels \
   --label_dtypes=$label_dtypes \
   --trainable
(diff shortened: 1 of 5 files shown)


sorensenjs:
It wasn't clear why we switched to a cached tf hub copy. I can do that as well if you want
just wasn't sure what the motivation was.

msushkov:
The hub model is cached to enable distributed training, which is needed to train on the P100 GPU. Otherwise some of the workers don't have access to the copy that's downloaded by another worker.

sorensenjs:
Done, PTAL

msushkov:
The model version is still 2 on the tf-hub page, best to keep it consistent. Just use gs://conversationai-models/resources/tfhub/kona-universal-encoder-2/1fb57c3ffe1a38479233ee9853ddd7a8ac8a8c47.

sorensenjs:
I'm using this one 
https://tfhub.dev/google/universal-sentence-encoder-large/3
so update the name (and moved the files) accordingly

msushkov:
Makes sense, thanks @sorensenjs!

msushkov:
LGTM

sorensenjs:
Changed again to put the paths in the shell scripts, giving the code a sensible default.
PTAL

experiments/tf_trainer/tf_hub_classifier/myrunlocal.sh
```diff
@@ -33,7 +33,7 @@ python -m tf_trainer.tf_hub_classifier.run \
   --train_path=$train_path \
   --validate_path=$valid_path \
   --model_dir="tf_hub_classifier_local_model_dir" \
-  --model_spec="https://tfhub.dev/google/universal-sentence-encoder-lite/2" \
+  --model_spec="gs://conversationai-models/resources/tfhub/universal-sentence-ecncoder-large-3/96e8f1d3d4d90ce86b2db128249eb8143a91db73" \
```
msushkov: Nitpicking here but typo in resource path

sorensenjs:
Fixed

PR #244, opened 2019-02-12
Save/export best checkpoint for model.

msushkov:


experiments/tf_trainer/common/model_trainer.py
```diff
@@ -287,16 +320,42 @@ def export(self, serving_input_fn, example_key_name=None):
...
     for checkpoint_path in checkpoints_to_export:
-      version = checkpoint_path.split('-')[-1]
-      estimator.export_savedmodel(
-          export_dir_base=os.path.join(self._model_dir(), version),
-          serving_input_receiver_fn=serving_input_fn,
-          checkpoint_path=checkpoint_path)
+      # If we are exporting the best checkpoint, see if the current checkpoint
+      # is the best.
+      if FLAGS.n_export == -1:
+        metrics = self._estimator.evaluate(self._dataset.validate_input_fn,
```
fprost: Can you find the metric value in the checkpoint? Instead of recomputing it?

experiments/tf_trainer/common/model_trainer.py
```diff
@@ -287,16 +320,42 @@ def export(self, serving_input_fn, example_key_name=None):
...
+    if not metrics_key and FLAGS.n_export == -1:
+      raise ValueError(
+        'Must provide value for `metrics_key` when exporting best checkpoint.')
+
     estimator = self._estimator
     if example_key_name:
       estimator = self._add_estimator_key(self._estimator, example_key_name)
-    
+
     checkpoints_to_export = self._get_list_checkpoint(FLAGS.n_export,
```
fprost: Can we insert all this logic in the function _get_list_checkpoint? I think this maps the logic of previous function

experiments/tf_trainer/common/model_trainer.py
msushkov: Done, using the tfevent file now.

experiments/tf_trainer/common/model_trainer.py
msushkov: Done.

msushkov:
Tried using tf.estimator.BestExporter and unfortunately ran into issues where multiple model versions were being exported still, and it was unclear which one was best.

fprost:
Interesting. Naive check: Have you set up the exports_to_keep to 1 (default seems 5)?

msushkov:
Yeah, tried that. Was getting 10 exported models for some reason.

msushkov:
Currently observing an issue with the BestExporter implementation: https://github.com/tensorflow/tensorflow/issues/25899.

This implementation works without using BestExporter, so using this currently. When the above issue is addressed we can switch to #246.

experiments/tf_trainer/common/model_trainer.py
```diff
@@ -233,30 +239,136 @@ def _add_estimator_key(self, estimator, example_key_name):
...
         os.path.join(model_dir, 'model.ckpt-*.index'))
     checkpoints = [x.replace('.index', '') for x in checkpoints]
     checkpoints = sorted(checkpoints, key=lambda x: int(x.split('-')[-1]))
 
     if n_export == 1:
       return [checkpoints[-1]]
+    elif n_export == -1:
+      # If we want only the best checkpoint, delete the others.
+      best_ckpt = self._get_best_checkpoint(checkpoints, metrics_key,
+        is_first_metric_better_fn)
```
fprost: nit: I believe that the indentation is not correct.
is_first_metric_fn should be aligned with checkpoints?

experiments/tf_trainer/common/model_trainer.py
```diff
@@ -233,30 +239,136 @@ def _add_estimator_key(self, estimator, example_key_name):
...
+      remaining_ckpts = set(checkpoints) - set([best_ckpt])
```
fprost: Could you help me to understand why you want to remove other checkpoints? follow-up: Should we do it for other situations as well?

experiments/tf_trainer/common/model_trainer.py
```diff
@@ -233,30 +239,136 @@ def _add_estimator_key(self, estimator, example_key_name):
...
+          in as arguments two numbers, returns true if first is better than
+          second. Default function says larger is better. Default value works for
+          AUC: higher is better.
 
     Returns:
-      List of checkpoint path.
+      List of checkpoint paths to export.
 
     If n_export==1, we take only the last checkpoint.
+    If n_expost==-1, we take the best checkpoint, according to `metrics_key` and
```
fprost: typo

experiments/tf_trainer/common/model_trainer.py
msushkov: Since we only care about the best checkpoint when n_export == -1 figured it might make sense to clean up the model directory. If this is something you think we should do for all cases I can put this code in export(). Best to be consistent, so either always delete all unexported checkpoints, or don't delete any. What do you think? Are there situations where the checkpoints (not the exported model) are useful?

msushkov: The only downside is extra storage for checkpoints that may not ever be needed.

fprost: It makes sense, let's delete it for every runs.

PR #248, opened 2019-02-28
Remove keras from our examples.

sorensenjs:


sorensenjs:
As an additional test before submitting  I'm running all of the models locally at least to 100 steps.

PR #252, opened 2019-03-28
Script to deploy list of models

fprost:
Utility to deploy a list of models.


fprost:
Made another commit. Now, there is no restriction to define a model with fewer than 7 versions.

Context: Currently, CMLE can not have more than 7 concurrent prediction jobs (quota).

Changes: If the model has more than 7, the Dataset frameworks "splits" the model into submodel with less than 7 versions.

model_evaluation/score_test_data.py
```diff
@@ -6,7 +22,7 @@
 import tensorflow as tf
 
 import input_fn_example
-from utils_export.dataset import Dataset, Model
+from utils_export.dataset import Dataset, Model, CMLE_QUOTA_PREDICTION
```
nthain: Why is this imported in this file if it's not used here?

fprost: Good point, removed quota here

PR #256, opened 2019-04-04
fix some comments and types

iislucas:
Also a step in exploring using pytype (for type-checking in python), and thinking about making a jsonline input class.

### DIFF
--- a/experiments/tf_trainer/common/tfrecord_input.py
+++ b/experiments/tf_trainer/common/tfrecord_input.py
@@ -6,7 +6,7 @@ from __future__ import print_function
 
 import multiprocessing
 import tensorflow as tf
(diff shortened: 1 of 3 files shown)


experiments/tf_trainer/common/tfrecord_input.py
```diff
@@ -70,7 +72,8 @@ def text_feature(self) -> str:
   def train_input_fn(self) -> types.FeatureAndLabelTensors:
     """input_fn for TF Estimators for training set.
 
-    Automatically repeats over input data forever.
+    Automatically repeats over input data forever. We define epoc limits in the
```
nthain: epoc -> epoch

PR #257, opened 2019-04-04
Script to convert data into bert-readable format

nthain:


experiments/tools/bert_tfrecord_converter.py
```diff
@@ -0,0 +1,106 @@
...
+tf.app.flags.DEFINE_string('filenames', None,
+                           'Comma separated list of filenames.')
+tf.app.flags.DEFINE_string('bert_url', 'https://tfhub.dev/google/bert_uncased_L-12_H-768_A-12/1', 'TF Hub URL for BERT Model')
+tf.app.flags.DEFINE_integer('max_sequence_length', 256,
+                            'Maximum sequence length of tokenized comment.')
+
+FLAGS = tf.app.flags.FLAGS
+
+def create_int_feature(values):
+      f = tf.train.Feature(int64_list=tf.train.Int64List(value=list(values)))
```
iislucas: Use 2 char intend consistently. :)

experiments/tools/bert_tfrecord_converter.py
```diff
@@ -0,0 +1,106 @@
...
+    in_filepath = '{}{}'.format(input_data_path, filename)
+    #TODO: Check if file exists, if not write new file
+    #TODO: Have the filename reflect the max_sequence_length and path reflect model
+    out_filepath = '{}{}'.format(output_data_path, filename)
+    record_iterator = tf.python_io.tf_record_iterator(path=in_filepath)
+    writer = tf.python_io.TFRecordWriter(out_filepath)
+    for ex_index, string_record in enumerate(record_iterator):
+      example = tf.train.Example()
+      example.ParseFromString(string_record)
+      text = example.features.feature[text_key].bytes_list.value[0]
```
iislucas: Confused about where text_key and label_key are defined? Did you want this to be a flag?

experiments/tools/bert_tfrecord_converter.py
nthain: Thanks! Done.

experiments/tools/bert_tfrecord_converter.py
nthain: Ah, good catch! Yes, they should be flags and are now.

PR #260, opened 2019-04-10
Adding cosine

fprost:
Adding the partial work on cosinus.


model_evaluation/Predict bias.ipynb
```diff
@@ -0,0 +1,370 @@
...
+   "outputs": [],
+   "source": [
+    "kernel_gender = np.transpose(checkpoint_utils.load_variable(MODEL_DIR_GENDER, 'final_layer/kernel'))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Compute cosinus."
```
nthain: Nit: I think we would say "Compute Cosine" (cosinus seems to be the French term)

model_evaluation/Predict bias.ipynb
```diff
@@ -0,0 +1,370 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 103,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "(33, 128)"
```
nthain: Is this the shape we expect for kernel_gender? Why not (2, 128) for just the two genders (male/female)?

model_evaluation/Predict bias.ipynb
```diff
@@ -0,0 +1,370 @@
...
+    }
+   ],
+   "source": [
+    "for i in range(33):\n",
+    "#     _bias = np.abs(cosine_similarity(kernel_gender_female, kernel_occupation[i].reshape(1, -1))) + \\\n",
+    "#                   np.abs(cosine_similarity(kernel_gender_male, kernel_occupation[i].reshape(1, -1)))\n",
+    "\n",
+    "    _bias = (cosine_similarity(kernel_gender_female - kernel_gender_male, kernel_occupation[i].reshape(1, -1)))\n",
+    "\n",
+    "#     _bias = cosine_similarity(direction_male, kernel_occupation[i].reshape(1, -1))\n",
```
nthain: Ah, I see what you mean when you say there are multiple ways to compute this cosine similarity. I'll try a few of them but the direction_male, direction_female intuitively feel like good places to start.

model_evaluation/Predict bias.ipynb
fprost: Done.

model_evaluation/Predict bias.ipynb
fprost: This is a great question.
This model had actually a bug: I had put n_classes=33 even if the only classes were {0,1}. Do you think that it might affect this analysis? I can retrain a model if needed.

Also, related to what we mentioned yesterday, we could discuss (2,128) vs (1,128). I was thinking that (2,128) makes our analysis a little harder, but is more generalizable to non-binary identity groups.
Let me know what you think

nthain: I think retraining with n_classes = 1 makes sense here. Having a unique "gender indicator" vector (basically a gender direction) will make the subsequent analysis easier. For extending to non-binary identity groups, this may still be the best path as we want an indicator direction for each member of the group (and it's not clear that the mean-subtraction techniques you use in the other notebook are better).

fprost: Sounds good. Code has been modified and new models will incorporate this change

fprost:
Pushing the notebook as is.

Once the models have been retrained, we will update the notebook to handle a (1,128) shape.

PR #272, opened 2019-07-08
generalize tfrecord converter

nthain:


experiments/tools/convert_csv_to_tfrecord.py
```diff
@@ -49,8 +65,13 @@ def main(argv):
 
   input_csv_path = FLAGS.input_csv_path
   output_tfrecord_path = FLAGS.output_tfrecord_path
+  column_names = FLAGS.column_list.split(',')
+  column_dtypes = FLAGS.dtype_list.split(',')
```
sorensenjs: assert len(column_names) == len(column_dtypes)

PR #276, opened 2019-07-26
Evaluation notebook for few-shot learning baselines

msushkov:


nthain:
I would at least get rid of the info and warning messages. There seem to be a ton of them and it makes the notebook quite hard to read.

msushkov:
Removed most outputs (except for the results themselves).

